PR #20, opened 2023-06-07
Frontend charm integration test run bugfix

mz2:
Fixes the sole frontend charm integration test scenario.

- Charm requires juju 3.x for Reasons, although the frontend charm does not explicitly set a requirement for juju 3.x (a known issue apparently that a charm requires a different `pylibjuju` depending on juju 2.x vs 3.x). I migrated to a fork of the `charmed-kubernetes/actions-operator` with which I could install more recent versions of juju and microk8s, in strict confinement (https://github.com/charmed-kubernetes/actions-operator/pull/57).
- The containerd template trick for GitHub image access when deploying the charm needed to be done also in the integration test workflow (this workflow now verifies the install prerequisite steps for local development too).
- Plot twist I realised is that the integration tests actually had not been run ever really on the workflow yet (they had run locally, but not ever really on GitHub correctly thus far).., and actually there was a minor regression caught by this simple test scenario that exists there now): when the frontend charm is deployed on its own, the charm should actually reach a maintenance status (it would error out after some changes that were made in response to the earlier _unit_ tests that didn't test the scenario where the charm is deployed on its own without the backend).
- Simplifies the relation handling logic: no dependence on `_stored`, API server hostname and port read from the relation's application data bag (see more at https://juju.is/docs/sdk/integration, in particular the section "Handling integration data from non-relation hooks").

Note: user `https://github.com/canonical-hw-certbot` was added as a collaborator to the project, and the token set in secrets is theirs (with only `packages:read` permission set).


mz2:
This has been superseded by https://github.com/canonical/test_observer/pull/54

PR #39, opened 2023-08-17
Update revision to be 20 and 13

nadzyah:
Update API and frontend revisions in the TF config for deployment

UPD: It turned out that both tox and flutter were updated yesterday, and we use the latest versions of both
So, I had to fix the actions in this PR as well


omar-selo (approved):

omar-selo:
Fixes for GH actions look good. It's worth noting that we now fix flutter version to 3.10.6. But that does have the benefit of things not breaking randomly

nadzyah:
It turned out that both tox and flutter were updated yesterday, and we use the latest versions of both
So, I had to fix the actions in this PR as well

PR #41, opened 2023-08-22
Fix api url not used by frontend charm

omar-selo:
Resolves (RTW-143)[https://warthogs.atlassian.net/browse/RTW-143]. Frankly, while these changes have fixed the issue for both my local setup and staging, I don't know what the issue is with the previous code. So it is possible that this issue may crop back again. Or it could be that something within these changes fixed it :shrug: 

Fyi, I've already deployed these changes into staging and observed that they work. As can be seen from [https://test-observer-staging.canonical.com/#/debs](https://test-observer-staging.canonical.com/#/debs). Currently, however snap board isn't working, although that's another bug.

### DIFF
--- a/frontend/charm/src/charm.py
+++ b/frontend/charm/src/charm.py
@@ -95,7 +95,7 @@ class TestObserverFrontendCharm(ops.CharmBase):
 
     def _on_rest_api_relation_broken(self, event):
         logger.debug("REST API relation broken")
-        self._update_layer_and_restart(event)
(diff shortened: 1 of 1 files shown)


nadzyah:
Just in case, the url in the description doesn't work. 
It should be [RTW-143](https://warthogs.atlassian.net/browse/RTW-143) `[RTW-143](https://warthogs.atlassian.net/browse/RTW-143)`

nadzyah (approved):

PR #44, opened 2023-09-29
Add initial sentry configuration for API

nadzyah:
This PR resolves https://warthogs.atlassian.net/browse/RTW-133 and adds support for Sentry in TO API.

Here is the project for TO API on sentry: https://sentry.is.canonical.com/canonical/test-observer-api/
The changes were done according to this doc since Sentry doesn't support FastAPI specifically: https://sentry.is.canonical.com/canonical/test-observer-api/getting-started/python/

Also, I've updated the link for docker in README since installing it as snap causes issues when it's running with LXD, and we use LXD local development env with our other projects.

UPD: I've also create TO staging API project in sentry: https://sentry.is.canonical.com/canonical/test-observer-api-staging/


nadzyah:
Just in case, I had to tell mypy to ignore `Cannot instantiate abstract class "init" with abstract attribute "__exit__"  [abstract]` error while initialising sentry client. It's a [known bug](https://github.com/getsentry/sentry-python/issues/1415) that was fixed in later version. But since our self-hosted Sentry instance doesn't support latest clients that include this bug fix, we had to go with this approach.

backend/test_observer/main.py
```diff
@@ -18,14 +18,22 @@
...
+import sentry_sdk
+
 
 from test_observer.controllers.router import router
 
-app = FastAPI()
 
+SENTRY_DSN = os.getenv("SENTRY_DSN")
+if SENTRY_DSN:
+    sentry_sdk.init(SENTRY_DSN)  # type: ignore
```
nadzyah, line 33: Just in case, I had to tell mypy to ignore `Cannot instantiate abstract class "init" with abstract attribute "__exit__"  [abstract]` error while initialising sentry client. It's a [known bug](https://github.com/getsentry/sentry-python/issues/1415) that was fixed in later version. But since our self-hosted Sentry instance doesn't support latest clients that include this bug fix, we had to go with this approach.

omar-selo:
I'm assuming this resolves the backend part and the frontend is coming in a separate PR?

nadzyah:
> I'm assuming this resolves the backend part and the frontend is coming in a separate PR?

Since API and frontend should be tracked separately on sentry (because they use different sentry clients), I've decided to create a PR for API first

backend/test_observer/main.py
```diff
@@ -18,14 +18,22 @@
...
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
+import sentry_sdk
+
 
 from test_observer.controllers.router import router
 
-app = FastAPI()
 
+SENTRY_DSN = os.getenv("SENTRY_DSN")
```
omar-selo, line 31: two things:
- we should probably have two sentry environments one for staging and another for production (I can only see production on sentry dashboard)
- we need to pass this environment variable through terraform -> charm config

omar-selo (changes requested):
Glad finally got Sentry.

Whenever I setup sentry before there was a way to catch all uncaught errors and send them to sentry. It would be great if we can setup this here.

backend/k8s/api.yaml
```diff
@@ -37,6 +37,9 @@ spec:
           image: localhost:32000/test-observer-api
           ports:
             - containerPort: 30000
+          env:
+            - name: SENTRY_DSN
+              value: "https://dd931d36e0c24681aaeed6abd312c896@sentry.is.canonical.com//66"
```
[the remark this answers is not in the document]
nadzyah: Don't we use k8s for prod as well?

backend/test_observer/main.py
nadzyah: I've noticed that other projects on sentry don't have anything for staging

backend/test_observer/controllers/router.py
```diff
@@ -41,3 +41,9 @@
 def root(db: Session = Depends(get_db)):
     db.execute(text("select 'test db connection'"))
     return "test observer api"
+
+
+@router.get("/sentry-debug")
+def trigger_error():
+    division_by_zero = 1 / 0
+    return division_by_zero
```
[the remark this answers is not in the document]
nadzyah: I've decided to keep it for future checks

backend/k8s/api.yaml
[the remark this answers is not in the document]
omar-selo: Nope we use charms and juju like hexr (but k8s charms, although that has nothing to do with these k8s yaml files)

backend/test_observer/main.py
omar-selo: That's ok it doesn't hurt to have staging and we could make use of it

nadzyah: I'm not sure if we care about errors on staging. And we also don't have production yet. Probably it makes sense to use it with staging for now and then for production only

omar-selo: I can see uses for staging. Imagine we deploy something on staging and find an issue. Sentry will send us frontend errors happening on the client. This way we can debug an issue on staging before deploying to production

plars: Agree, it can sometimes be useful in staging also

backend/k8s/api.yaml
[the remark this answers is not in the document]
plars: yeah, we should remove this. It's not going to be useful outside of whatever development environment it's set up in, and would just be confusing. However we might need to add some support in the terraform plan here to load that DSN into the environment, right?

plars (changes requested):
Agree with Omar + question about if we should add support for injecting it through a variable that we pass to terraform

terraform/test-observer.tf
```diff
@@ -25,6 +25,13 @@ variable "external_ingress_hostname" {
   default     = "canonical.com"
 }
 
+locals {
+  sentry_dsn_map = {
+    production  = "https://dd931d36e0c24681aaeed6abd312c896@sentry.is.canonical.com//66"
+    staging     = "https://84a48d05b2444e47a7fa176b577bf85a@sentry.is.canonical.com//68",
+  }
```
omar-selo, line 32: needs a value for development or else `local.sentry_dsn_map[var.environment]` will probably fail (that could just be an empty string)

backend/charm/config.yaml
```diff
@@ -8,3 +8,6 @@ options:
     default: test-observer-api
     description: Public hostname for the service.
     type: string
+  sentry_dsn:
+    description: value for SENTRY_DSN that is used to send errors to sentry
+    type: string
```
omar-selo, line 13: small note, we could have a default value of "" so that it's easier to deploy locally through `juju deploy test-observer-api`

omar-selo (changes requested):
Good work, just mainly a note on development value of sentry dsn

omar-selo (approved):
LGTM!

plars (changes requested):
Small change to suggest, which I hope actually makes it a little easier. I don't think we want to hardcode those dsn values, just make them a variable. Otherwise, looking great!

plars (approved):
After discussing this a bit more on MM, I think it would be easiest to just do this for now, since it wouldn't require setting these DSNs in the environment. I still think that it would be nice to have this as a configurable option in the future, but we could easily change that after we get the ability to deploy all of this in a more modular way. So +1

PR #45, opened 2023-10-02
Add terraform support for specifying a source range to allow

plars:


terraform/test-observer.tf
```diff
@@ -25,6 +25,12 @@ variable "external_ingress_hostname" {
   default     = "canonical.com"
 }
 
+variable "nginx_ingress_integrator_charm_whitelist_source_range" {
+  description = "Allowed client IP source ranges. The value is a comma separated list of CIDRs."
+  type        = string
+  default     = ""
+}
```
omar-selo, line 32: so once we make the firewall changes to allow PS5 deployments to be accessed without a VPN, we should also set this value to empty string for Test Observer right?

omar-selo (approved):

PR #46, opened 2023-10-02
Inject sentry_dsn as a string and expand vars in it

plars:
Without this, I was getting the following error when trying to deploy locally:
```
│   80:     sentry_dsn = {local.sentry_dsn_map[var.environment]}                
│                                                                               
│ Expected an equals sign ("=") to mark the beginning of the attribute value.   
╵ 
```


plars:
one more small fix for _app_environment

omar-selo (approved):
Oh I should've tested it before approving last PR

nadzyah (approved):
Oops, thank you so much for fixing this!

PR #47, opened 2023-10-03
Add sentry client for TO frontend

nadzyah:
This PR resolves https://warthogs.atlassian.net/browse/RTW-133 and adds sentry reporting to TO frontend.

Since for our sentry instance (v9.2.1) there's no flutter or dart client, I've decided to use the oldest available `sentry_flutter` package that is also compatible with our flutter version. The client was configured according to this doc: https://pub.dev/packages/sentry_flutter/versions/5.0.0

Sentry projects were created for both [production](https://sentry.is.canonical.com/canonical/test-observer-frontend/) and [staging](https://sentry.is.canonical.com/canonical/test-observer-frontend-staging/) frontend instances.

### DIFF
--- a/frontend/charm/config.yaml
+++ b/frontend/charm/config.yaml
@@ -15,3 +15,8 @@ options:
     description: Scheme for accessing Test Observer API (used to construct URLs for API requests)
     default: "https://"
     type: string
+
+  sentry_dsn:
+    description: Sentry DSN name for sending errors to sentry server
+    default: ""
+    type: string
--- a/frontend/charm/src/charm.py
+++ b/frontend/charm/src/charm.py
@@ -97,7 +97,7 @@ class TestObserverFrontendCharm(ops.CharmBase):
         logger.debug("REST API relation broken")
         self._handle_no_api_relation()
 
-    def nginx_config(self, base_uri: str) -> str:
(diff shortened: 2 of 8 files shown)


nadzyah:
Unfortunately, sentry dashboard doesn't represent any received events for some reason. However, we've done everything from the client side and, as you can see, the server successfully accepted 185 requests from the client. 
image

It seems to be an issue on sentry server side, likely because its version is too old for flutter support.

omar-selo:
> Unfortunately, sentry dashboard doesn't represent any received events for some reason. 
> ...
> It seems to be an issue on sentry server side, likely because its version is too old for flutter support.

I'm a bit confused does that mean sentry is currently not working on the frontend? or were we able to send errors from frontend and can see them on the dashboard?

nadzyah:
> I'm a bit confused does that mean sentry is currently not working on the frontend? or were we able to send errors from frontend and can see them on the dashboard?

I think that sentry doesn't know how to represent the events but accepts them anyway.

nadzyah:
> Perhaps that makes it unusable. I wonder if using their API directly would work as otherwise I fail to see the usefulness of it on frontend.

Agree, but I don't know what is more reasonable, to ask IS for an upgrade or to write our custom client

nadzyah:
I'm closing it for now since we don't know when we get flutter support on our Sentry server

PR #48, opened 2023-10-04
Update upload-charm action to v2.4 to support node16 on github

plars:
Github is deprecating node12 so apparently this is needed soon? now?


nadzyah (approved):

plars:
this might (probably) still need the pull-image thing that omar mentioned. I hadn't seen that when I pushed this. But this piece will be needed whenever github deprecates node 12 in any case

PR #49, opened 2023-10-16
Update flutter and dependencies major versions

omar-selo:
Flutter version got bumped up. If we used the latest flutter the code wouldn't work cause some packages we're depending on need updating to be compatible. So I figured I'd upgrade the major version of everything. This ofc involved a few changes to the code.

### DIFF
--- a/.github/workflows/test_frontend.yml
+++ b/.github/workflows/test_frontend.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: subosito/flutter-action@v2
         with:
-          flutter-version: '3.10.6'
+          flutter-version: '3.13.0'
           channel: 'stable'
       - run: flutter pub get
       - run: flutter pub run build_runner build
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get update
 RUN apt-get install -y curl git wget unzip libgconf-2-4 gdb libstdc++6 libglu1-mesa fonts-droid-fallback lib32stdc++6 python3
 
 RUN git clone https://github.com/flutter/flutter.git /opt/flutter && \
-    cd /opt/flutter && git checkout '3.10.6' && \
+    cd /opt/flutter && git checkout '3.13.0' && \
     /opt/flutter/bin/flutter doctor && \
     /opt/flutter/bin/flutter config --enable-web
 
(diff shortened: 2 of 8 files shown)


.github/workflows/test_frontend.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: subosito/flutter-action@v2
         with:
-          flutter-version: '3.10.6'
+          flutter-version: '3.13.0'
```
p-gentili, line 17: Since we're bumping, why not the latest? 3.13.7

frontend/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apt-get update
 RUN apt-get install -y curl git wget unzip libgconf-2-4 gdb libstdc++6 libglu1-mesa fonts-droid-fallback lib32stdc++6 python3
 
 RUN git clone https://github.com/flutter/flutter.git /opt/flutter && \
-    cd /opt/flutter && git checkout '3.10.6' && \
+    cd /opt/flutter && git checkout '3.13.0' && \
```
p-gentili, line 9: ditto

frontend/lib/models/test_execution.dart
```diff
@@ -57,7 +57,7 @@ enum TestExecutionStatus {
       case inProgress:
         return const Icon(YaruIcons.refresh, size: size);
       case passed:
-        return const Icon(YaruIcons.ok, color: YaruColors.success, size: size);
+        return Icon(YaruIcons.ok, color: YaruColors.light.success, size: size);
```
p-gentili, line 60: dumb Q: are we Ok with not checking the theme here?

p-gentili:
Ops, already merged :)

frontend/lib/models/test_execution.dart
omar-selo: Both dark and light themes have the same color for success. But mainly, our dashboard doesn't support different themes. i.e. you'll see the same colors irrelevant of what you pick

.github/workflows/test_frontend.yml
omar-selo: Now I understand what's going on. I'm using flutter snap. I thought a snap refresh would mean I get the latest flutter version. Turns out no, a refresh just refreshes the snap but not flutter itself. To upgrade flutter I'd need to run flutter upgrade. This makes sense as we wouldn't want snap automatically upgrading flutter versions.

PR #50, opened 2023-10-16
Fix/frontend api connection on dev

omar-selo:
It seems that front-end right now can't connect to back-end when running flutter normally outside of charms. This is because it depends on the charm's nginx swapping the place holder back-end url with the correct one. So this PR sets the placeholder url to be localhost so that if the url is not changed through the charm, it would work fine on development environment.

### DIFF
--- a/backend/skaffold.yaml
+++ b/backend/skaffold.yaml
@@ -1,7 +1,7 @@
 apiVersion: skaffold/v4beta4
 kind: Config
 metadata:
-  name: test-observer-frontend
+  name: test-observer-api
 build:
   artifacts:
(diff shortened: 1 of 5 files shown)


plars (approved):
nice, you tested this locally in your scaffold setup I assume? Did you also happen to try local deployment via the charm to ensure that works properly still?

omar-selo:
> nice, you tested this locally in your scaffold setup I assume? Did you also happen to try local deployment via the charm to ensure that works properly still?

Good point. Just tested out the charms locally on my machine and they work too.

PR #52, opened 2023-10-23
Artefact links

omar-selo:
Resolves [RTW-148](https://warthogs.atlassian.net/browse/RTW-148)

Basically, this PR is about making artefacts link-able. As right now there is no links to artefacts. This PR adds the link using the path `/family/:artefactId`

Sadly, this PR is much bigger than I would've liked. This is because it involved some changes and refactors both in frontend and backend. Main issue was that when I made it possible to link directly to artefacts, it no longer became possible to pass artefact data from parent widget (the stage column). Instead, the artefact dialog has to fetch artefact information on it's own. Additionally, the artefact dialog could no longer get stages information and should either fetch that too or in this PR's case just hard code the stages on the frontend. I figured there is no need to make model changes on the backend.

[RTW-148]: https://warthogs.atlassian.net/browse/RTW-148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/tests/controllers/artefacts/test_artefacts.py
```diff
@@ -17,13 +17,53 @@
...
 
+def test_get_latest_artefacts_by_family(db_session: Session, test_client: TestClient):
+    """Should only get latest artefacts and only ones that belong to given family"""
+    relevant_artefact = create_artefact(db_session, "edge", version="2")
+
+    old_timestamp = relevant_artefact.created_at - timedelta(days=1)
+    create_artefact(db_session, "edge", created_at=old_timestamp, version="1")
+    create_artefact(db_session, "proposed")
+
+    response = test_client.get("/v1/artefacts?family=snap")
```
nadzyah, line 37: Is there a difference between `/v1/artefacts?family=snap` and `/v1/artefacts/?family=snap`?

frontend/lib/models/stage_name.dart
```diff
@@ -0,0 +1,20 @@
...
+        StageName.candidate,
+        StageName.stable,
+      ];
+    case FamilyName.deb:
+      return [
+        StageName.proposed,
+        StageName.updates,
+      ];
+  }
+}
```
nadzyah, line 20: shouldn't this be dynamic and depend on backend response?

frontend/lib/routing.dart
```diff
@@ -31,4 +58,16 @@ final appRouter = GoRouter(
...
+
+  static FamilyName familyFromContext(BuildContext context) {
+    final route = GoRouterState.of(context).fullPath!;
+
+    if (route.startsWith(snaps)) {
+      return FamilyName.snap;
+    } else if (route.startsWith(debs)) {
+      return FamilyName.deb;
+    } else {
+      throw Exception('Unknown route: $route');
```
nadzyah, line 70: if-else statements look fine for now since we have only 2 families. but isn't it better to use switch-case because in future we may have more families (image, for instance)?

nadzyah (commented):
Thanks for your changes! Sorry if some questions are ignorant.

backend/tests/controllers/artefacts/test_artefacts.py
omar-selo: As far as I know they're both used interchangeably

frontend/lib/routing.dart
omar-selo: I couldn't use switch statement with `startsWith` method

backend/tests/controllers/artefacts/test_artefacts.py
nadzyah: Shouldn't we pick one option and use it consistently?

frontend/lib/routing.dart
nadzyah: ah, definitely! no problem in this case

backend/tests/controllers/artefacts/test_artefacts.py
omar-selo: But I noticed that I can pass query parameters as an object. So I changed it to that

frontend/lib/models/stage_name.dart
omar-selo: We don't expect these values to change. Yes we may add images later, but in that PR we can extend these values. The reason for this is that we avoid having to create endpoints for families and stages. And more importantly we avoid having to fetch them and show a spinner on the frontend before they arrive. i.e. it's just simpler. But if you still feel strongly that we should fetch families and stages I could try to work on it in a separate PR cause it will probably another big change.

nadzyah (approved):

PR #53, opened 2023-10-25
Fix frontend api request block due to mixed content

omar-selo:
Right now [TO staging fails to fetch data from API](https://test-observer-staging.canonical.com/#/snaps) because the API request is [blocked by the browser](https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content/How_to_fix_website_with_mixed_content). After digging into what's going on, I found that frontend tries to request artefacts through the following url `https://test-observer-api-staging.canonical.com/v1/artefacts?family=snap`. The api then responds with a 307 redirect to `http://test-observer-api-staging.canonical.com/v1/artefacts/?family=snap` because fastapi redirects request to match trailing slash of endpoint. But the issue is that the redirect switches from https scheme to http. The browser detects this and blocks the request.

Note that this problem doesn't exist on local development because local development doesn't use a tls certificate. So all requests are http. Hence there is no mixed content shenanigans.

### DIFF
--- a/backend/test_observer/controllers/application/version.py
+++ b/backend/test_observer/controllers/application/version.py
@@ -1,10 +1,11 @@
-from importlib.metadata import version, PackageNotFoundError
+from importlib.metadata import PackageNotFoundError, version
+
 from fastapi import APIRouter
 
 router = APIRouter()
 
 
-@router.get("/")
+@router.get("")
 async def get_version():
     try:
         return {"version": version(__package__)}
--- a/backend/test_observer/controllers/artefacts/artefacts.py
+++ b/backend/test_observer/controllers/artefacts/artefacts.py
@@ -29,7 +29,7 @@ from .models import ArtefactBuildDTO, ArtefactDTO
 router = APIRouter()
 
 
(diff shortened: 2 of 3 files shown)


plars (approved):
I'm still not sure I understand why the initial request to https://test-observer-api-staging.canonical.com/v1/artefacts/?family=snap was blocked by the browser, but I understand why it blocked it after trying http instead.

Otherwise, this looks ok to me. The only comment I have for future (not that big of a concern, don't worry about fixing it here) is that it looks like there were a lot of unrelated cleanup changes to order of imports, etc. Try to at least put them in a separate commit if it's not too much trouble, because it helps with readability. But there weren't too many here so it was ok.

omar-selo:
@plars the initial request was to https://test-observer-api-staging.canonical.com/v1/artefacts?family=snap and it was redirected to http://test-observer-api-staging.canonical.com/v1/artefacts/?family=snap. i.e. a slash was added after artefacts and turned https to http.

Good point regarding import ordering. Didn't notice that. Perhaps the way ruff formats is different from isort. So when we switched to ruff it changed things. But thanks for the tip, I'll keep it in mind.

PR #54, opened 2023-10-26
Frontend charm integration tests fix with a self hosted runner

mz2:
Integration tests for the frontend charm now actually work also on GitHub!

The workflow looks to be stable, with a dozen passes witnessed today.

It does however take ~10mins to run and I figured we would therefore not start running it on every commit to every branch (6+ mins of the time is taken by the microk8s setup -- we can drop this in a big way with custom images used for the CI runs when that becomes available). It runs on:
- all branches once a PR review has been posted for the branch for which there is a PR open
- all branches with a PR which touch resources under `./frontend/charm` (the integration tests can capture failures also in resources outside that directory, hence the rule above also).
- on the main branch
- on version tagged commits
- manually via workflow dispatch

The workflow was refactored to reuse two now introduced reusable composite actions in https://github.com/canonical/certification-github-workflows (the [first PR to that repo](https://github.com/canonical/certification-github-workflows/pull/2) should be reviewed and merged first before this goes in).

**NOTE!** When squashing this into main, please edit the commit message heavily down in volume since this contains a few months worth of desperate debugging attempts of these integration tests that continued to work locally just fine but failed to be executed specifically on GitHub up until now. I have simplified the version history with some squashing but since there are several merge commits there I didn't want to do anything more drastic than that.


plars (changes requested):
A few questions below

.github/workflows/frontend_charm_integration_tests.yml
```diff
@@ -1,24 +1,52 @@
...
+            [registry."docker.io"]
+              mirrors = ["https://github-runner-dockerhub-cache.canonical.com:5000"]
       - name: Setup LXD
         uses: canonical/setup-lxd@main
         with:
-          channel: 5.13/stable
+          channel: 5.19/stable
       - name: Setup operator environment
-        uses: charmed-kubernetes/actions-operator@main
+        uses: mz2/actions-operator@4fcb55b99e2a0f69a5910870d3639d203aeb87aa
```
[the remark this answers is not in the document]
mz2: Because otherwise I can't force strictly confined versions of juju and microk8s to be used.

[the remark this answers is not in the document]
mz2: I have posted a PR with my changes, indeed should be able to merge back.

For clarity this branch when in a mergeable state will be the first point at which these charm integration tests actually behave.

frontend/charm/metadata.yaml
```diff
@@ -6,7 +6,7 @@ description: |
   Test Observer Frontend offers a dashboard for presenting test results served by the test-observer backend service.
 
 assumes:
-  - juju >= 2.9
+  - juju >= 3.1
```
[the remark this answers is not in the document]
mz2: (I doubt it'll be a problem since with this present at 2.9, juju 3.1 and 3.2 have run the application successfully in local dev envs)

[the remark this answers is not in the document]
mz2: I have changed this back to 2.9 and it all works (well, it builds and integration and unit tests pass anyway) still.

.github/workflows/frontend_charm_integration_tests.yml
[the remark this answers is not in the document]
mz2: Plot twist: turns out my tweaks here were unnecessary because snap ignores `--classic` for a strictly confined snap: https://github.com/charmed-kubernetes/actions-operator/pull/57 --> I have changed this back to the way it was before.

.github/workflows/publish_charms.yml
```diff
@@ -0,0 +1,32 @@
...
+    name: Publish Charm
+    needs: integration-test
+    runs-on: [self-hosted, linux]
+    if: ${{ github.ref_name == 'main' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install charmcraft
+        run: sudo snap install charmcraft --classic
+      - name: Fetch Tested Charm
```
mz2, line 18:
```suggestion
      - name: Fetch the tested charm
```

frontend/charm/requirements.txt
```diff
@@ -1 +1,2 @@
-ops >= 2.2.0
+ops >= 2.8.0
+macaroonbakery==1.3.2
```
mz2, line 2: Added as a workaround to https://github.com/canonical/test_observer/actions/runs/7142353669/job/19451476954 as instructed at https://chat.canonical.com/canonical/pl/9xzxm61rgtry5d7sfx9xz7ryyc

frontend/charm/tox.ini
```diff
@@ -66,7 +66,7 @@ commands =
 description = Run integration tests
 deps =
     pytest
-    juju
+    juju<=3.2
```
mz2, line 69: This is intentionally `<= 3.2` as opposed to for example `>= 2.9` as is set in the charm metadata: the integration test environment needs to be (for reasons of various version incompatibilities) 3.x.

.github/workflows/frontend_charm_integration_tests.yml
```diff
@@ -1,45 +1,26 @@
...
-        uses: canonical/setup-lxd@main
-        with:
-          channel: 5.13/stable
-      - name: Setup operator environment
-        uses: charmed-kubernetes/actions-operator@main
-        with:
-          provider: microk8s
+        uses: actions/checkout@v4
+      - name: Set up microk8s
+        uses: canonical/certification-github-workflows/.github/actions/microk8s-setup@microk8s-setup-action
```
mz2, line 19: The setup steps turned out to be more extensive on self hosted runners, and since this is all repetitive boilerplate, moved to a reusable composite action. Same story below re: the `archive-charm-testing-artifacts`

terraform/test-observer.tf
```diff
@@ -1,7 +1,7 @@
 terraform {
   required_providers {
     juju = {
-      version = "~> 0.7.0"
```
mz2, line 4: The changes below showed up as necessary when the provider was updated: the `series` (and `base`) are nonsensical for k8s charms => removed.

mz2: This change is not strictly speaking necessary for the charm integration tests (though it was my intention to also test out in succession also that the terraform plan continues to be able to spin a new environment up that stabilises in an all-ready state).

.github/workflows/publish_api.yml
```diff
@@ -8,7 +8,7 @@ env:
 
 jobs:
   build-and-push-image:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, linux, xlarge]
```
omar-selo, line 11: this change will require setting up proxy so that images can be pushed to `ghcr`. Unless self hosted runners have been updated and don't require this anymore?

.github/workflows/publish_frontend.yml
```diff
@@ -8,7 +8,7 @@ env:
 
 jobs:
   build-and-push-frontend-image:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, linux]
```
omar-selo, line 11: Same here

frontend/charm/requirements.txt
omar-selo, line 2: can you add a comment to remind us why this is needed?

terraform/test-observer.tf
omar-selo: can we do this on PS5 though? as in can we use tf provider for juju version `0.10.0`?

omar-selo (commented):
It's amazing that you got it to work! finally this has been annoying for a long time

.github/workflows/frontend_charm_integration_tests.yml
```diff
@@ -1,45 +1,34 @@
-name: Run charm tests
+name: Run charm integration tests
 on:
+  pull_request_review:
+    types: [submitted]
```
omar-selo, line 4: perhaps we don't need this as it triggers whenever someone reviews the PR

terraform/test-observer.tf
mz2: I think this should indeed Just Work, in that the next time you deploy using this plan, it will upgrade the provider to the latest version that is described here. Deploying to staging will confirm that safely, I'd say.

.github/workflows/publish_api.yml
mz2: Indeed no longer needed. :partying_face:

mz2: ... however, both here and in the frontend image publishing workflow it is necessary to still configure a mirror for docker.io such as not to hit rate limits. Done that in a follow-up commit now.

.github/workflows/publish_frontend.yml
mz2: Proxying no longer needed to access ghcr.io, but docker.io rate limits could still be hit --> mirror configuration added.

.github/workflows/frontend_charm_integration_tests.yml
mz2: Turned out I was missing an `if` condition as per https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved (and as implemented in the metabox tests from github.com/canonical/checkbox where I copied this from)

mz2:
Superseded by https://github.com/canonical/test_observer/pull/80

PR #55, opened 2023-10-26
Moves to self-hosted runners

mz2:
Moves from GitHub hosted runners to self-hosted runners.

As you can see, the changes in the repo are straightforward. The "large" runners are faster than GitHub hosted runners (2x to 3x faster juju & microk8s bootstrap time in the case of the integration tests I reopened efforts on in a different branch). I intentionally did not specify the `large` where I thought we don't require it (and builds indeed seem to pass happily), since we get workers then assigned from a larger pool.

To get the repo working with these runners requires the repository to conform to these rules: https://github.com/canonical/repo-policy-compliance -- the thing that might bite at first is [signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) being required for all the commits from the point this is merged (but there are at least no open PRs at the moment, so strikes as a good time to introduce this).

### DIFF
--- a/.github/workflows/frontend_charm_analysis.yml
+++ b/.github/workflows/frontend_charm_analysis.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   lint-report:
     name: Lint and format report
-    runs-on: ubuntu-22.04
+    runs-on: [self-hosted, linux, large]
     steps:
       - uses: actions/checkout@v3
       - name: Install tox
@@ -23,7 +23,7 @@ jobs:
 
   static-analysis:
     name: Static analysis
-    runs-on: ubuntu-22.04
+    runs-on: [self-hosted, linux]
     steps:
       - uses: actions/checkout@v3
       - name: Install tox
@@ -36,7 +36,7 @@ jobs:
 
   unit-tests-with-coverage:
     name: Unit tests
-    runs-on: ubuntu-22.04
+    runs-on: [self-hosted, linux]
     steps:
       - uses: actions/checkout@v3
       - name: Install tox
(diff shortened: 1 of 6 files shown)


omar-selo (approved):
Nice! Btw I currently have a branch open, does that mean I need to sign the commits I already made on that branch?

plars (approved):
lgtm, the only question I have is why some specify the large tag and others do not. IIRC, large was the smallest worker type that we have on these right? but even if it wasn't, things like "lint-report" was marked as large but "build-and-push-image" was not. Would it make more sense to drop the large tag from all of them and then specify xlarge or two-xlarge only if needed for larger things?

mz2:
> Btw I currently have a branch open, does that mean I need to sign the commits I already made on that branch?

Yes indeed. We went through this same rollercoaster with Checkbox earlier this week, and there's some instructions here regarding how you can get them signed: https://github.com/canonical/checkbox/blob/main/CONTRIBUTING.md#signed-commits-required

mz2:
> gtm, the only question I have is why some specify the large tag and others do not. IIRC, large was the smallest worker type that we have on these right? but even if it wasn't, things like "lint-report" was marked as large but "build-and-push-image" was not. Would it make more sense to drop the large tag from all of them and then specify xlarge or two-xlarge only if needed for larger things?

I didn't specify anything where I thought there's no specific requirement (small or large will do).

I didn't specify anything above `large` since I didn't actually spot they existed 🤦 However, build times seem sufficient with the `large` and the speedup from that to the integration test scenario bootstrapping was already substantial. I'd propose iterating on this on the branch where the charm integration tests are fixed.

PR #57, opened 2023-10-31
Fix/ci missing docker buildx

omar-selo:
It seems that self hosted runners do not include `buildx` unlike GH runners, so docker build phase is failing on both [api](https://github.com/canonical/test_observer/actions/runs/6671002387/job/18132005227) and [frontend](https://github.com/canonical/test_observer/actions/runs/6671002401/job/18132005236). This PR hopefully fixes that. But I wasn't able to test it, as the usual hack of running actions on none main branch doesn't work anymore due to [repo-policy-compliance](https://github.com/canonical/repo-policy-compliance) requiring CI to only run on protected branches.

### DIFF
--- a/.github/workflows/publish_api.yml
+++ b/.github/workflows/publish_api.yml
@@ -17,6 +17,9 @@ jobs:
       IMAGE_NAME: ${{ github.repository }}/api
 
     steps:
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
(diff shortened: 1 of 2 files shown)


omar-selo:
Actually now that I think about it, CI should be able to run on this branch. So I need to figure out why it's [failing policy check](https://github.com/canonical/test_observer/actions/runs/6704421809/job/18216796382).

PR #60, opened 2023-11-02
Fix ORM model and add alembic check

omar-selo:
It seems that I had modified Artefacts table through a migration, but forgot to make the same change on the ORM. The specific change was creating two uniqueness constraints one for snaps and another for debs, as they can't be grouped into one. I detect this issue when I tried to generate a migration automatically through alembic. To prevent this issue from happening again, I added a CI check that will fail if alembic detects a difference between ORM models and what the migrations actually do.

### DIFF
--- a/backend/test_observer/data_access/models.py
+++ b/backend/test_observer/data_access/models.py
@@ -109,9 +109,8 @@ class Artefact(Base):
     status: Mapped[ArtefactStatus | None]
 
     __table_args__ = (
-        UniqueConstraint(
-            "name", "version", "track", "series", "repo", name="unique_artefact"
-        ),
+        UniqueConstraint("name", "version", "track", name="unique_snap"),
+        UniqueConstraint("name", "version", "series", "repo", name="unique_deb"),
     )
 
     def __repr__(self) -> str:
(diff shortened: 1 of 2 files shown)


backend/test_observer/data_access/models.py
```diff
@@ -109,9 +109,8 @@ class Artefact(Base):
     status: Mapped[ArtefactStatus | None]
 
     __table_args__ = (
-        UniqueConstraint(
-            "name", "version", "track", "series", "repo", name="unique_artefact"
-        ),
+        UniqueConstraint("name", "version", "track", name="unique_snap"),
```
nadzyah, line 112: could you please tell me what this "name" field mean? how does the DB know that it's related only to snaps?

omar-selo: "name" is the name of the Artefact. Unless you mean `name="unique_snap"` which means the name of the constraint on the database.

The way this works all depends on how PostgreSQL handles null values in a unique constraint. Basically, if any of the values in the constraint is null then postgresql would essentially say this whole record is unique, even if other the values already exists. For example, if two artefacts had the same `name` and `version` but had null track then postgresql doesn't complain. But if they also had the same track then postgresql would not allow it. That means the the `unique_snap` constraint doesn't do anything when the artefact is a deb because track would be null. Instead, only the `unique_deb` constraint stops duplicates. Similarly, the `unique_deb` constraint doesn't do anything when the artefact is a snap because both series and repo are null. Instead, only the `unique_snap` constraints stops duplicate.

nadzyah (approved):

PR #61, opened 2023-11-09
Fetch testexecution status from c3

omar-selo:
Resolves [RTW-135](https://warthogs.atlassian.net/browse/RTW-135). and [RTW-147](https://warthogs.atlassian.net/browse/RTW-147).

# Frontend changes (screenshots)

Test Observer Frontend

Screenshot from 2023-11-28 14-15-31
Screenshot from 2023-11-28 14-15-41

# Implementation details

## RTW-135 Changes:
- Figure out test execution status by fetching submission report from C3 at the time that a user fetches artefact builds (which embed test executions)
- Remove test execution status from data model as it's now fetched on the fly rather than stored. This is because the previously stored value is actually incorrect. Jenkins doesn't actually know the status of a test execution so it can't send it to TO. Additionally, TO can't get the status of the test execution when it gets c3_link because the execution might not have finished yet
- Add `UNKNOWN` test execution status and remove unused `NOT_STARTED` and `NOT_TESTED` (will add back if we need them later)
- Made small updates to frontend just show question mark for the new `UNKNOWN` status
- Pass C3 api credentials through terraform and charm to application

## RTW-147 Changes:
- Backend changes:
    - New TestResult data model is introduced
    - Test cases and execution details are fetched from C3 and parsed into the Report objects
    - All test cases are exposed on the `/artefact/<ID>/builds` endpoint
- Frontend changes:
    - New TestResult data model is introduced
    - Two new components are added to render the Test cases: GroupTestResultView and _TestResultView 

[RTW-135]: https://warthogs.atlassian.net/browse/RTW-135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[RTW-147]: https://warthogs.atlassian.net/browse/RTW-147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/test_observer/controllers/artefacts/logic.py
```diff
@@ -0,0 +1,119 @@
...
+
+class TestExecutionStatusLogic:
+    NO_C3_LINK = TestExecutionStatus.IN_PROGRESS
+    NO_STATUS_ID_IN_C3_LINK = TestExecutionStatus.UNKNOWN
+    SUBMISSION_STATUS_NOT_FOUND = TestExecutionStatus.UNKNOWN
+    SUBMISSION_STATUS_IS_FAIL = TestExecutionStatus.FAILED
+    NO_REPORT_ID_IN_SUBMISSION_STATUS = TestExecutionStatus.UNKNOWN
+    REPORT_NOT_FOUND = TestExecutionStatus.UNKNOWN
+    NO_FAILED_TESTS = TestExecutionStatus.PASSED
+    SOME_FAILED_TESTS = TestExecutionStatus.FAILED
```
nadzyah, line 28: should it be an enum?

backend/test_observer/controllers/artefacts/logic.py
```diff
@@ -0,0 +1,119 @@
...
+    if submission_status.status == SubmissionProcessingStatus.FAIL:
+        return TestExecutionStatusLogic.SUBMISSION_STATUS_IS_FAIL
+    if (report_id := submission_status.report_id) is None:
+        return TestExecutionStatusLogic.NO_REPORT_ID_IN_SUBMISSION_STATUS
+    if (report := reports.get(report_id)) is None:
+        return TestExecutionStatusLogic.REPORT_NOT_FOUND
+    if report.failed_test_count == 0:
+        return TestExecutionStatusLogic.NO_FAILED_TESTS
+    else:
+        return TestExecutionStatusLogic.SOME_FAILED_TESTS
```
nadzyah, line 112: you can remove else after return statement. however, if you want to keep else in this if-else sequence, I guess it's logical to use `elif` instead of just `if` stating from the 2nd if

backend/test_observer/external_apis/c3/c3.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def get_submissions_statuses(
+        self, ids: Iterable[int]
+    ) -> dict[int, SubmissionStatus]:
+        str_ids = [str(status_id) for status_id in ids]
+        response = self._authenticate_and_send(
+            Request(
+                method="GET",
+                url="https://certification.canonical.com/api/v2/submissions/status/",
+                params={"id__in": ",".join(str_ids), "limit": len(str_ids)},
```
nadzyah, line 27: right now limit doesn't work. should this PR depend on this change (to add `limit` param for v2 API) in c3?

backend/test_observer/controllers/artefacts/logic.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+
+        build_dto = ArtefactBuildDTO(
+            id=build.id,
+            architecture=build.architecture,
+            revision=build.revision,
+            test_executions=test_executions_dto,
+        )
+
+        yield build_dto
```
andrejvelichkovski, line 89: Maybe we can directly yield this data instead of assigning it to a variable.

backend/test_observer/controllers/artefacts/logic.py
```diff
@@ -0,0 +1,119 @@
...
+        .order_by(ArtefactBuild.architecture, ArtefactBuild.revision.desc())
+        .options(
+            joinedload(ArtefactBuild.test_executions).joinedload(
+                TestExecution.environment
+            )
+        )
+        .all()
+    )
+
+    return orm_builds
```
andrejvelichkovski, line 45: Similarly here, maybe we can directly return this query

backend/test_observer/external_apis/c3/c3.py
```diff
@@ -0,0 +1,74 @@
...
+            logger.warning(response.text)
+            return {}
+
+    def _authenticate_and_send(self, request: Request) -> Response:
+        prepared_request = request.prepare()
+        prepared_request.headers["Authorization"] = f"Bearer {self._bearer_token}"
+        session = requests.session()
+        response = session.send(prepared_request)
+        if response.status_code in (401, 403):
+            self._authenticate()
```
andrejvelichkovski, line 61: It seems like this if-condition checks whether the authorization token is still valid and refreshes it, should we maybe extract this into a helper function `_refresh_and_resend_request()` or a similar name to make it a little bit more clear on why we are resending the request?

andrejvelichkovski (commented):
Overall the changes look good to me, I left a few style comments in the code, but nothing related to the logic of the code. Let's wait for a review by Nadzeya too, as she is probably more knowledgeable for the code than me.

backend/test_observer/external_apis/c3/c3.py
```diff
@@ -0,0 +1,48 @@
...
+        self.client_id = client_id
+        self.client_secret = client_secret
+
+        self.bearer_token = ""
+
+    def get_submissions_statuses(self, ids: Iterable[int]):
+        session = requests.session()
+        request = Request(
+            method="GET",
+            url="https://certification.canonical.com/api/v2/submissions/status/",
```
andrejvelichkovski: Should we update the url to be more dynamic, as we might want to work in C3 and TO development at the same time?

backend/test_observer/controllers/artefacts/logic.py
```diff
@@ -89,6 +93,23 @@ def construct_dto_builds(
         yield build_dto
 
 
+def _derive_test_results(
```
omar-selo, line 96: perhaps to avoid repeating some of the logic in `_derive_test_execution_status` we can instead have one function return both `status` and `test_results` as a tuple. We need a nice name for it though

backend/test_observer/controllers/artefacts/models.py
```diff
@@ -58,6 +60,7 @@ class TestExecutionDTO(BaseModel):
     c3_link: str | None
     environment: EnvironmentDTO
     status: TestExecutionStatus
+    test_results: list[TestResult] | None
```
omar-selo, line 63: I sort of understand that there is a distinction between `[]` and `None` here. As you could argue that `None` means we didn't get a report yet, while `[]` means the report has no tests. But as I see the code actually always returns `[]` so this is not really the case here. Besides it's easier to deal with a list always rather than an optional. So perhaps we can drop `None` type

frontend/lib/models/test_execution.dart
```diff
@@ -16,6 +17,7 @@ class TestExecution with _$TestExecution {
     @JsonKey(name: 'c3_link') required String? c3Link,
     required TestExecutionStatus status,
     required Environment environment,
+    @JsonKey(name: 'test_results') required List<TestResult>? testResults,
```
omar-selo, line 20: if we drop optional `None` type on backend then we don't need to handle nullable here

backend/tests/external_apis/test_c3.py
```diff
@@ -67,13 +111,34 @@ def test_get_submissions_statuses(
...
+    }
+    requests_mock.get(
+        f"https://certification.canonical.com/api/v2/reports/summary/?id__in={c3_api_response['id']}&limit=1",
+        request_headers={"Authorization": f"Bearer {bearer_token}"},
+        json={"results": [c3_api_response]},
+    )
+
     requests_mock.get(
-        f"https://certification.canonical.com/api/v2/reports/summary/?id__in={report['id']}&limit=1",
+        f"https://certification.canonical.com/api/v2/reports/summary/{c3_api_response['id']}/",
```
omar-selo, line 127: does this endpoint get called? I thought just the previous mocked endpoint will be called

backend/tests/external_apis/test_c3.py
```diff
@@ -67,13 +111,34 @@ def test_get_submissions_statuses(
...
-        json={"results": [report]},
+        json={"results": [c3_api_response]},
     )
 
-    reports = c3.get_reports([report["id"]])
+    # We ignore this typing error, it happens because of the
+    # "testresult_set" assigned to empty list in the c3_api_response
+    # dict, and the Python Typing system cannot be sure that the "id"
+    # field is int
+    reports = c3.get_reports([c3_api_response["id"]])  # type: ignore
```
omar-selo, line 136: you could use [TypedDict](https://docs.python.org/3.10/library/typing.html#typing.TypedDict) here

backend/tests/external_apis/test_c3.py
```diff
@@ -67,13 +111,34 @@ def test_get_submissions_statuses(
...
 
-    assert reports == {report["id"]: Report(**report)}
+    expected_report = Report(
+        id=237670,
+        failed_test_count=0,
+        test_count=0,
+        test_results=[],
```
omar-selo, line 142: would be nice to have a test case that has a test result

frontend/lib/models/test_results.dart
```diff
@@ -0,0 +1,55 @@
...
+part 'test_results.freezed.dart';
+part 'test_results.g.dart';
+
+@freezed
+class TestResult with _$TestResult {
+  const factory TestResult({
+    required int id,
+    required String name,
+    required String type,
+    required TestResultStatus status,
```
omar-selo, line 15: Nice, I see you used an enum for test result status (I guess we could use one in the backend too). How about `type` can it be an enum?

frontend/lib/ui/dashboard/artefact_dialog.dart
```diff
@@ -274,6 +276,23 @@ class _TestExecutionView extends StatelessWidget {
     final jenkinsLink = testExecution.jenkinsLink;
     final c3Link = testExecution.c3Link;
 
+    // Split the test result in separate categories based on Status (PASS/FAIL/SKIP)
+    final Map<TestResultStatus, List<TestResult>> testResultGroups = {};
```
omar-selo, line 280: you could use [groupBy](https://api.flutter.dev/flutter/package-collection_collection/groupBy.html)

frontend/lib/ui/dashboard/testresults_components.dart
```diff
@@ -0,0 +1,106 @@
...
+          const SizedBox(width: Spacing.level4),
+          Text(
+            testResultStatus.name,
+            style: Theme.of(context).textTheme.titleLarge,
+          ),
+          const Spacer(),
+          Text(
+            groupTestResults.length.toString(),
+            style: Theme.of(context).textTheme.titleLarge,
+          ),
```
omar-selo, line 32: I'm not sure if there were any discussions on this already, but it seems more clear to me to have the number of test results on the left not right. But this is open for discussion

frontend/lib/ui/dashboard/testresults_components.dart
```diff
@@ -0,0 +1,106 @@
...
+              )
+              .toList(),
+        ),
+      ),
+    );
+  }
+}
+
+class _TestResultView extends StatelessWidget {
+  const _TestResultView({required this.testResult});
```
omar-selo, line 51: can you update the screenshot on the PR to show this widget. As I don't know what it should look like

omar-selo (commented):
Great job picking up some `FastAPI` and `Flutter`, not sure if you had prior experience with them. I just had some small comments. Also I have a question regarding the design. Not sure if there were discussions on how it should look like on the frontend. Because it's different from the [initial design on figma](https://www.figma.com/file/378oTyxZdHQeqEpSo9Jh8E/Test-Observer-Dashboard-Design?type=design&node-id=535-715&mode=design&t=J0aaHrqW0IuMiOWt-0).

backend/test_observer/external_apis/c3/c3.py
omar-selo: Good point. It doesn't sound good for dev deployment to call C3 production. But I don't want this to break in development either. So perhaps we could actually create another class altogether that implements the same public methods as this `C3` class. Except that class would just return say empty objects (cause it's for development). Then we'd use FastAPI's dependency injection to use the appropriate class depending on the environment. This would also mean that we need to pass the environment to the backend application.

backend/migrations/versions/2023_11_03_0951-3f23c3dc0d74_add_unknown_to_test_execution_status.py
```diff
@@ -0,0 +1,44 @@
...
+    op.add_column(
+        "test_execution",
+        sa.Column(
+            "status",
+            sa.Enum(
+                "Not Started",
+                "In Progress",
+                "Passed",
+                "Failed",
+                "Not Tested",
```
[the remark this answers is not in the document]
omar-selo: I actually dropped the `status` column from `test_execution` table. As this is now computed on the fly from C3

backend/test_observer/external_apis/c3/c3.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def _authenticate(self) -> None:
+        response = requests.post(
+            "https://certification.canonical.com/oauth2/token/",
+            auth=(self._client_id, self._client_secret),
+            data={"grant_type": "client_credentials"},
+        )
+
+        if response.ok:
+            self._bearer_token = response.json().get("access_token", "")
```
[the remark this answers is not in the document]
omar-selo: If we want to keep track of exactly one token then this will over complicate the code. Especially if in the future we deploy multiple TO backend nodes. When I tested it I noticed that we can have as many tokens as we want and it would still work. So I figured to simplify things we just create a token every session. FastAPI's dependency injection will use the same dependency (i.e. same token) on the same session. But from my understanding there is no way to make it use the same dependency across sessions. At least no easy way to do so

backend/test_observer/controllers/artefacts/logic.py
omar-selo: That's what I did initially. But I started getting linting errors or mypy (I forgot which one) that stated you can't have an enum whose value type is an enum too

frontend/lib/models/test_results.dart
andrejvelichkovski: I firstly thought that this `type` field has some meaning, but I then noticed that it always has the value `test`. I think it would be best to remove it as it doesn't seem to have any meaning behind it and we don't use it in any of our views.

frontend/lib/ui/dashboard/testresults_components.dart
andrejvelichkovski: I have now added a few screenshots. However, since this is first time I'm using Flutter, I couldn't figure out how to align all this stuff nicely to the left, and it ends up in the centre of the dialog.

backend/test_observer/external_apis/c3/models.py
```diff
@@ -14,6 +14,17 @@ class SubmissionStatus(BaseModel):
     report_id: int | None
 
 
+class TestResult(BaseModel):
+    id: int
+    name: str
+    type: str
+    status: str
```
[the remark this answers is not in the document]
andrejvelichkovski: I think we should remove the `type` parameter, it always has the value `test` which is not valuable. For status, it is string field in HEXR, but seems to have only three values: fail, pass, skip. However, looking at the historic database I found out there are some test results that have more types: fail, pass, skip, uninitiated, unresolved, unsupported, untested. Let me double check if we still plan to support these.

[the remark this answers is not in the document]
andrejvelichkovski: Update: the other four types haven't been used since 2015 and 2018, I think it is safe to ignore them.

backend/test_observer/controllers/artefacts/models.py
andrejvelichkovski: I was initially having the same view on having `None` type, but then decided it is much simpler to just return empty list if no results are available. I forgot to remove the optional `None` type here though. Thanks for flagging it!

backend/test_observer/external_apis/c3/models.py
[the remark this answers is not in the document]
omar-selo: Good find. btw are there nulls? and can we make sure that they will not be used? 

Also regarding `status` dropping it makes sense to me too

frontend/lib/ui/dashboard/testresults_components.dart
```diff
@@ -0,0 +1,106 @@
...
+                        style: Theme.of(context).textTheme.titleMedium,
+                      ),
+                      Text(
+                        testResult.ioLog,
+                      ),
+                    ],
+                  )
+                : const SizedBox.shrink(),
+          ],
+        ),
```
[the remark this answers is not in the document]
andrejvelichkovski: Thanks for suggesting this! I'm still picking up Flutter and didn't know about these features. I tried using what you suggested and it works better, but it still gets rendered to the centre if the text width does not cover the full width of the widget.

Screenshot from 2023-11-28 10-33-12

backend/tests/external_apis/test_c3.py
andrejvelichkovski: Apologies for leaving this here, I used this endpoint before updating C3 to return the test results in the first endpoint as well.

frontend/lib/ui/dashboard/testresults_components.dart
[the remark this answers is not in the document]
omar-selo: hmm, try adding [textAlign](https://api.flutter.dev/flutter/widgets/RichText/textAlign.html) property to the `RichText` widget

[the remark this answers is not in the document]
andrejvelichkovski: It didn't change anything, but I manage to resolve it by wrapping the RichText widget inside Align widget where I specified alignment: Alignment.topLeft

backend/tests/controllers/artefacts/test_artefacts.py
```diff
@@ -202,6 +203,7 @@ def test_correct_test_execution_status(
                         "name": environment.name,
                         "architecture": environment.architecture,
                     },
+                    "test_results": [],
```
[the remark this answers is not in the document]
andrejvelichkovski: I have now added such test case

backend/test_observer/controllers/artefacts/logic.py
andrejvelichkovski: We can do this, I left it in two separate ones to simplify it, but we can combine it into something like: `_derive_test_execution_details` which would fetch both

backend/test_observer/external_apis/c3/c3.py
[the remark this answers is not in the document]
nadzyah: Does it mean that you create a token each time you make a request to C3 to fetch the submissions data? I'm afraid that this approach will make it harder to track all these tokens on C3 side.

nadzyah (changes requested):

nadzyah:
> @nadzyah not sure how it works on Django, but typically tokens are not stored or managed on server. Typically it just decrypts the token and reads the TTL and user from it. At least that's how [JWT](https://jwt.io/) work.

Can we store it encrypted on the server? I just don't want TO to generate thousands of tokens that are used just once but actually are valid for 30 days

omar-selo:
> Can we store it encrypted on the server?

I'll try to investigate further how to do this. But I still don't see the disadvantages of TO doing so. So if it ends up being over complicated, I will not opt for it

andrejvelichkovski:
I think this is exactly what it does right now, the C3API class has an attribute `_bearer_token` which only updates when it expires but it's used for all requests that come through this instance.

andrejvelichkovski:
I don't mind changing what we represent, however, I think these are two different kinds of data. In the first picture, it is an IO Log of an individual test case, while on the second picture we have a test summary for all test cases that run on the machine. The information in the second screenshot is still available on the Test Observer, just in slightly different format.

andrejvelichkovski:
Ah sorry, missed your point, we can surely update the front-end to represent this in an code-like syntax.

PR #63, opened 2023-11-22
Enable publishing the images with self hosted runners again

mz2:
This re-introduces the self hosted runners, since after a discussion with @jdkandersson I was left with the impression that the self hosted runners _should_ be able to reach ghcr.io through the firewall (the actual issue we look to be hitting does not get us as far as attempting to connect to ghcr.io since the base image is actually coming from docker.io).

Corresponding Mattermost thread tracking the problem:
https://chat.canonical.com/canonical/pl/jz5ds1ko5pboirjau9dp5pdado

### DIFF
--- a/.github/workflows/publish_api.yml
+++ b/.github/workflows/publish_api.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: ["main"]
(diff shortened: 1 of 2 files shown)


mz2:
Doesn't seem relevant anymore?

PR #66, opened 2023-12-01
Remove sys.exit() that breaks the promote endpoint

andrejvelichkovski:
This PR should fix the issue with artefacts not being promoted on the Test Observer boards.

The issue with it was the command `sys.exit()` which doesn't get caught as an exception and causes Internal Server Error which couldn't be handled. 

The snap that caused this was "cascade-configuration" which cannot be found on the channel '16'.

## Test

Two screenshots of the boards (one before the fix, one after the fix). We can notice two snaps are now in the stable board

Screenshot from 2023-12-01 11-00-00
Screenshot from 2023-12-01 11-00-15


omar-selo:
So in this case if we drop `sys.exit(1)` the code will continue. But would that be an issue? I imagine it would break somewhere?

omar-selo (approved):
:crossed_fingers: hope it works!

PR #70, opened 2023-12-05
Set naming conventions for database constraints

omar-selo:
Right now database constraints (primary keys, foreign keys, indexes, etc) don't have a fixed naming convention in SQLAlchemy. This results in Alembic automatically creating these constraints without an actual name. That makes the database choose the name. This makes constraints a bit [more challenging to work with on alembic](https://alembic.sqlalchemy.org/en/latest/naming.html). This PR basically adds these naming conventions to SQLAlchemy and renames those that are not following these conventions.


andrejvelichkovski (approved):
Looks good to me!

PR #72, opened 2023-12-05
Receive report and set testexecution status

omar-selo:
This implements part of [RTW-196](https://warthogs.atlassian.net/browse/RTW-196)

Changes:
- Add `PUT /v1/test-executions/end-test` endpoint that receives a report (from C3)
- Make `ci_link` unique (technically I just wanted to add an index but I don't see it wouldn't be unique)
- Set the test execution status depending on the received report

To be done in later PR:
- Store individual TestResults and return them through API then display them on frontend

[RTW-196]: https://warthogs.atlassian.net/browse/RTW-196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/test_observer/controllers/test_executions/test_executions.py
```diff
@@ -99,6 +104,26 @@ def start_test_execution(
...
+    if test_execution is None:
+        raise HTTPException(status_code=404, detail="Related TestExecution not found")
+
+    failed = any(r.status == C3TestResultStatus.FAIL for r in request.test_results)
+
+    if failed:
+        test_execution.status = TestExecutionStatus.FAILED
+    else:
+        test_execution.status = TestExecutionStatus.PASSED
+    db.commit()
```
nadzyah, line 124: sorry if it's a dumb question, but is it okay that we don't return any response here?

backend/test_observer/data_access/models.py
```diff
@@ -210,7 +210,7 @@ class TestExecution(Base):
 
     __tablename__ = "test_execution"
 
-    ci_link: Mapped[str | None] = mapped_column(String(200), nullable=True)
+    ci_link: Mapped[str | None] = mapped_column(String(200), nullable=True, unique=True)
     c3_link: Mapped[str | None] = mapped_column(String(200), nullable=True)
```
nadzyah, line 214: shouldn't C3 link be unique as well?

andrejvelichkovski (approved):
Approving the changes as overall it looks good to me! I only left one comment with a concern I had about the `category` field.

backend/test_observer/data_access/models.py
omar-selo: Yeah that's true. But I added uniqueness to ci_link as more of a means to add an index. Since we need to query test executions by ci_link. Currently, we don't need to query or filter by c3_link so there is no benefit of an index being there. So it feels like it would just add some extra cost on the db without benefit (as indexes are not free). Unless we want to enforce uniqueness which I technically don't see too much of an issue violating here (if it happens somehow). wdyt?

backend/test_observer/controllers/test_executions/models.py
```diff
@@ -58,6 +59,27 @@ def validate_required_fields(self) -> "StartTestExecutionRequest":
...
+    PASS = "pass"
+    FAIL = "fail"
+    SKIP = "skip"
+
+
+class C3TestResult(BaseModel):
+    id: int
+    name: str
+    status: C3TestResultStatus
+    category: str
```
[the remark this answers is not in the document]
omar-selo: If it's easy to parse we could include with the changes on C3 side? If not then I guess I prefer we remove it as we don't know when we'll add it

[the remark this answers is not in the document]
andrejvelichkovski: It should be a small change, I can try to include it with the current changes.

backend/test_observer/controllers/test_executions/test_executions.py
omar-selo: By default if we don't return anything at least the response would be 200 if it works

backend/test_observer/data_access/models.py
nadzyah: are there potential errors we can avoid by enforcing c3_link to be unique?

omar-selo: Hmm, right now we get this link from Jenkins (but we can later get it from C3). So suppose there was an error somewhere and two test executions were sent the same c3_link. If we have uniqueness constraint enforced then only one of them will have it, but that might not necessarily be the right one. The other will be empty. On the other hand, if we don't have uniqueness enforced then both will have the same link, which is an error too. I guess in all cases it's incorrect. So in that sense perhaps enforcing uniqueness isn't all that valuable

PR #76, opened 2023-12-07
Store individual test results

omar-selo:
This PR continues on [RTW-196](https://warthogs.atlassian.net/browse/RTW-196). It basically will store individual `TestResults` on TO so that TO can serve it to the frontend later on.

[RTW-196]: https://warthogs.atlassian.net/browse/RTW-196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/test_observer/controllers/test_executions/test_executions.py
+++ b/backend/test_observer/controllers/test_executions/test_executions.py
@@ -22,6 +22,10 @@
 from fastapi import APIRouter, Depends, HTTPException
 from sqlalchemy.orm import Session
 
+from test_observer.controllers.test_executions.logic import (
+    compute_test_execution_status,
+    parse_c3_test_results,
+)
 from test_observer.data_access.models import (
     Artefact,
     ArtefactBuild,
@@ -34,7 +38,6 @@ from test_observer.data_access.repository import get_or_create
 from test_observer.data_access.setup import get_db
 
 from .models import (
-    C3TestResultStatus,
     EndTestExecutionRequest,
     StartTestExecutionRequest,
     TestExecutionsPatchRequest,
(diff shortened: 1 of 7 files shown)


backend/test_observer/controllers/test_executions/test_executions.py
```diff
@@ -130,12 +138,10 @@ def end_test_execution(request: EndTestExecutionRequest, db: Session = Depends(g
...
         raise HTTPException(status_code=404, detail="Related TestExecution not found")
 
-    failed = any(r.status == C3TestResultStatus.FAIL for r in request.test_results)
-
-    if failed:
-        test_execution.status = TestExecutionStatus.FAILED
-    else:
-        test_execution.status = TestExecutionStatus.PASSED
+    test_cases, test_results = parse_c3_test_results(request.test_results)
+    db.add_all(test_cases)
```
andrejvelichkovski, line 142: I noticed the unique constraint on `name`, but was wondering what would be the behaviour of this command be in case some of the test cases indeed exist in the database, does it complain or it only adds the new ones.

andrejvelichkovski (dismissed):
Looks good! Can we just confirm that it works well when trying to add a test case that already exists in the database, that it doesn't complain or throw errors?

omar-selo:
@andrejvelichkovski great catch, turned out it wouldn't work if we reused test cases :sweat_smile:

backend/test_observer/controllers/test_executions/test_executions.py
omar-selo: Yup you were indeed right!

andrejvelichkovski (approved):
This looks much better now! Thanks for following up on it.

PR #77, opened 2023-12-07
Add test execution review functionality

andrejvelichkovski:
This PR introduces a new backend endpoint for reviewing a test execution.

Our decision was that we should have review process for both Test Execution and the whole Artefact. The artefact signoff is already implemented, but we still need to have similar implementation for Test Execution.

## Implementation

As part of this PR there are a few different updates:
* New model updates (added `review_status` and `review_comment`)
* Updated existing PATCH endpoint for Test Execution

### Database migrations

As part of this update, we need to apply one more migration to the database.

## Testing

Automated tests were added and a set of manual tests were performed to verify the behaviour.


backend/test_observer/data_access/models_enums.py
```diff
@@ -35,6 +35,15 @@ class TestExecutionStatus(Enum):
     NOT_TESTED = "NOT_TESTED"
 
 
+class TestExecutionReviewStatus(str, Enum):
+    UNDECIDED = "UNDECIDED"
+    MARKED_AS_FAILED = "MARKED_AS_FAILED"
+    APPROVED_INCONSISTENT_TEST_DEF = "APPROVED_INCONSISTENT_TEST_DEF"
+    APPROVED_UNSTABLE_PHSYICAL_INFRA = "APPROVED_UNSTABLE_PHSYICAL_INFRA"
+    APPROVED_FAULTY_HARDWARE = "APPROVED_FAULTY_HARDWARE"
+    APPROVED_GENERIC = "APPROVED_GENERIC"
```
omar-selo, line 44: Should we perhaps have instead of `APPROVED_GENERIC` have both `APPROVED_AUTOMATIC` and `APPROVED_OTHER`?

backend/test_observer/controllers/test_executions/test_executions.py
```diff
@@ -160,3 +161,22 @@ def patch_test_execution(
...
 
     db.commit()
+
+
+@router.patch("/{id}/review")
+def review_test_execution(
+    id: int,
+    request: TestExecutionsReviewPatchRequest,
+    db: Session = Depends(get_db),
+):
```
omar-selo, line 171: why not extend regular `PATCH /test-executions/:id` endpoint?

backend/test_observer/controllers/test_executions/models.py
```diff
@@ -81,6 +85,30 @@ class EndTestExecutionRequest(BaseModel):
 
 
 class TestExecutionsPatchRequest(BaseModel):
-    c3_link: HttpUrl | None = None
-    ci_link: HttpUrl | None = None
-    status: TestExecutionStatus | None = None
+    c3_link: HttpUrl | None
+    ci_link: HttpUrl | None
+    status: TestExecutionStatus | None
```
omar-selo, line 90: any reason for removing optional parameters in the old patch request?

omar-selo (changes requested):
Thanks for the work! Two main points are whether we need an additional endpoint, and whether we need a list of review statuses.

backend/test_observer/data_access/models_enums.py
```diff
@@ -35,6 +35,15 @@ class TestExecutionStatus(Enum):
     NOT_TESTED = "NOT_TESTED"
 
 
+class TestExecutionReviewStatus(str, Enum):
+    UNDECIDED = "UNDECIDED"
+    MARKED_AS_FAILED = "MARKED_AS_FAILED"
+    APPROVED_INCONSISTENT_TEST_DEF = "APPROVED_INCONSISTENT_TEST_DEF"
```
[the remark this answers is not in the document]
andrejvelichkovski: I wanted to shorten the `DEFINITION` word as this enum name was becoming too long

backend/test_observer/controllers/test_executions/models.py
andrejvelichkovski: I think git detected this as a change because I haven't merged main. Will make sure to check it after merging main.

backend/test_observer/data_access/models_enums.py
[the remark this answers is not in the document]
omar-selo: Ooh ok that wasn't obvious for me. What's wrong with just `APPROVED_INCONSISTENT_TEST`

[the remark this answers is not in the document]
andrejvelichkovski: That makes sense as well, will update it.

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def upgrade() -> None:
+    op.execute(
+        "CREATE TYPE testexecutionreviewstatus AS "
+        "ENUM( "
+        "'UNDECIDED', "
+        "'MARKED_AS_FAILED', "
+        "'APPROVED_INCONSISTENT_TEST_DEF', "
+        "'APPROVED_UNSTABLE_PHSYICAL_INFRA', "
```
mz2, line 25:
```suggestion
        "'APPROVED_UNSTABLE_PHYSICAL_INFRA', "
```

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
```diff
@@ -0,0 +1,43 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "CREATE TYPE testexecutionreviewstatus AS "
+        "ENUM( "
+        "'UNDECIDED', "
+        "'MARKED_AS_FAILED', "
```
mz2, line 23:
```suggestion
        "'FAILED', "
```

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
```diff
@@ -0,0 +1,43 @@
...
+# revision identifiers, used by Alembic.
+revision = "e39a45a38d1f"
+down_revision = "871eec26dc90"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "CREATE TYPE testexecutionreviewstatus AS "
```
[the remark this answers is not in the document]
mz2: (Are there other review statuses expected?)

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
```diff
@@ -0,0 +1,43 @@
...
+        "ENUM( "
+        "'UNDECIDED', "
```
mz2, line 22:
```suggestion
        "'UNKNOWN', "
```

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
```diff
@@ -0,0 +1,43 @@
...
+        "'MARKED_AS_FAILED', "
+        "'APPROVED_INCONSISTENT_TEST_DEF', "
+        "'APPROVED_UNSTABLE_PHSYICAL_INFRA', "
+        "'APPROVED_FAULTY_HARDWARE', "
+        "'APPROVED_GENERIC' "
```
mz2, line 27:
```suggestion
        "'APPROVED_TESTS_PASSED "
```

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
```diff
@@ -0,0 +1,43 @@
...
+        ")"
+    )
+    op.execute(
+        "ALTER TABLE test_execution ADD COLUMN "
+        "review_status testexecutionreviewstatus[] NOT NULL "
```
mz2, line 32:
```suggestion
        "review_statuses review_status[] NOT NULL "
```

backend/test_observer/data_access/models.py
```diff
@@ -225,6 +235,11 @@ class TestExecution(Base):
     status: Mapped[TestExecutionStatus] = mapped_column(
         default=TestExecutionStatus.NOT_STARTED
     )
+    review_status: Mapped[list[TestExecutionReviewStatus]] = mapped_column(
+        ARRAY(Enum(TestExecutionReviewStatus)),
+        default=[TestExecutionReviewStatus.UNDECIDED],
+    )
+    review_comment: Mapped[str | None]
```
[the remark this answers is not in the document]
mz2: That makes sense. If SQLAlchemy allows for check constraints, there are some further validations to do later: certain statuses should require a non-empty comment for example.

backend/test_observer/data_access/models.py
```diff
@@ -225,6 +235,11 @@ class TestExecution(Base):
     status: Mapped[TestExecutionStatus] = mapped_column(
         default=TestExecutionStatus.NOT_STARTED
     )
+    review_status: Mapped[list[TestExecutionReviewStatus]] = mapped_column(
+        ARRAY(Enum(TestExecutionReviewStatus)),
+        default=[TestExecutionReviewStatus.UNDECIDED],
+    )
```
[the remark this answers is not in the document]
mz2: It's because we established that multiple reasons for approving might apply, even per environment.

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
mz2: Actually you don't need this state at all since it's to be used as a list (an empty list would be an "unknown" state)

backend/test_observer/data_access/models.py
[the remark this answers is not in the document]
omar-selo: Ooh, that's a good point. But from a design perspective, I'm not sure this is clear or clean. It feels like we want to have comments/reasons/approval per TestResult but at the level of TestExecution somehow. I'd prefer if either we go full on to the level of TestResult or just stay at TestExecution or even just Artefacts with comments. I'm trying to think of some conceptual way to simplify this, as it seems over complicated right now in my head (both code and simply understanding from user perspective)

backend/test_observer/data_access/models.py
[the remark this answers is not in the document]
omar-selo: Yup this should be doable

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
omar-selo: Or just `REJECTED`? Cause right now `TestExecution` already has a `status` field that can be `FAILED`. So having another `review_status` field that can also be `FAILED` would be confusing

andrejvelichkovski: I also `REJECTED` would be better alternative to distinguish from other status types

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
andrejvelichkovski: Good point, will follow-up with removing it.

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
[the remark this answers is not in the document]
andrejvelichkovski: According to the Spreadsheet document I created, these were the ones added there.

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
andrejvelichkovski: Unfortunately, the enum name must match the enum class name in Python, according to SQLAlchemy

backend/test_observer/data_access/models.py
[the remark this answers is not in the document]
andrejvelichkovski: We were also discussing this before with @mz2, however the other alternative was to have approval on the level of different Test Cases, which might add more complexity to the whole workflow, especially if more than a few test cases fail because of the same error, you would need to go through all of them and mark them as `APPROVED`, so we decided to keep some balance and have approval on the level of Test Execution, similarly to what we had in Trello.

backend/test_observer/controllers/test_executions/test_executions.py
andrejvelichkovski: I have switched to using this endpoint now. My idea was to have a separate endpoint for this to make the other one cleaner, but I guess it doesn't make sense to add a new endpoint for something that we already have.

backend/test_observer/data_access/models.py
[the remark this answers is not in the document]
omar-selo: I'm thinking of complexity as how much explanation needs to be given to a user or developer prior to working on the feature at hand. By that definition I feel that if we take the approach of a list of approval states, then we have a lot of explanation to do for both developers and users. On the other hand the approaches below require much less explanation if any:
1. Approval per test result. If test result model had a review state it's sort of clear. The UI is also clear as we will have review button next to each test result that failed.
2. Approval per test execution but with only one approval state. Any explanation is in comments. Yes we can do less automation but this is very flexible and simple (automation on the list approach is hard too)
3. Approval per artefact but with only one approval state. Any explanation is in comments. From my understanding, this is what we currently have on trello. Given that we're replacing trello we can start with same functionality then improve once we have a better design and purpose

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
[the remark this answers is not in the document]
mz2: I meant this in context of renaming this type as just "review_status"

backend/test_observer/data_access/models.py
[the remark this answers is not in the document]
andrejvelichkovski: Following a short discussion with @mz2 on MM, there are a few reasons why we should go with the current workflow:
* Having a list of approval reasons to pick will let us improve our process overall. For example, we can easily find devices that lead to failing test cases very often, and use this knowledge to make decisions on what to do with them. Another type of analysis we might be able to do is to find tests that fail often and focus on improving them.
* Since we are implementing a new workflow, it might be better to focus on making it good now, instead of reworking it later once again.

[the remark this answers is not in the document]
omar-selo: Looking at both these points, I then see that option 1 is better. It gives us more control to analyze and report exactly what happened and what we approved. We can have an enum of approval reasons that we select only one option from (not a list)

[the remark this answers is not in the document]
mz2: The option (1) would indeed be the most accurate solution, it was something I was going to request initially (we also discussed it last week), but it leads to a lot of UX friction: if there are 100 failed tests because the hardware is faulty, for example, you have a 100 buttons to press. I was trying to propose this current level of granularity as a compromise between the user friction and the reporting standard.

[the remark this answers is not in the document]
mz2: > we will still have to answer whether there is approval per test execution too?

There should IMO downstream of this per artifact review choice be an option to "ignore" or "dismiss" a certain specific test result and to make that dismissal persistent, so that a test when executed on a specific device would not prevent signing off automatically even if it fails. This is going to be helpful for use of Test Observer with a) prerelease use cases (known issues are acceptable for an edge kernel, for example) and b) avoiding manual busywork when there are known-to-be-problematic test cases + test devices combos in the pool.

When implemented, that will need to come with similarly an explicit set of options for explaining why (and those may be the same review states we are introducing here).

Basically, I think the per execution level detail will _also_ be needed, just not yet and it will have a different supporting use.

backend/migrations/versions/2023_12_07_1451-e39a45a38d1f_add_test_execution_review_status_and_.py
mz2: That makes sense.

backend/test_observer/controllers/artefacts/models.py
```diff
@@ -52,6 +56,8 @@ class TestExecutionDTO(BaseModel):
     c3_link: str | None
     environment: EnvironmentDTO
     status: TestExecutionStatus
+    review_status: list[TestExecutionReviewDecision]
```
omar-selo, line 59: making `review_status` a set would be more representative and it would be great to add a comment here reminding us why this can have multiple values

backend/test_observer/controllers/test_executions/models.py
```diff
@@ -103,6 +104,21 @@ class TestExecutionsPatchRequest(BaseModel):
...
+        if self.review_status is None:
+            return self
+
+        if len(self.review_status) <= 1:
+            return self
+
+        for review_status in self.review_status:
+            if review_status in (TestExecutionReviewDecision.REJECTED,):
+                raise ValueError("Test execution can either be rejected or approved")
+        return self
```
omar-selo, line 121: this can be a [`field_validator`](https://docs.pydantic.dev/latest/concepts/validators/#field-validators) as it's only one field.

backend/test_observer/controllers/test_executions/models.py
```diff
@@ -103,6 +104,21 @@ class TestExecutionsPatchRequest(BaseModel):
...
+    def validate_review_status(self) -> "TestExecutionsPatchRequest":
+        if self.review_status is None:
+            return self
+
+        if len(self.review_status) <= 1:
+            return self
+
+        for review_status in self.review_status:
+            if review_status in (TestExecutionReviewDecision.REJECTED,):
+                raise ValueError("Test execution can either be rejected or approved")
```
omar-selo, line 120: perhaps it's easier to read something like:
```
if TestExecutionReviewDecision.REJECTED in self.review_status:
    raise ValueError("Test execution can either be rejected or approved")
```

backend/test_observer/controllers/test_executions/test_executions.py
omar-selo, line 1: We should probably also handle the case of test execution reruns. Perhaps if a rerun does occur then it might be appropriate to empty both `review_status` and `review_comment`

backend/test_observer/data_access/models.py
```diff
@@ -229,6 +239,11 @@ class TestExecution(Base):
     status: Mapped[TestExecutionStatus] = mapped_column(
         default=TestExecutionStatus.NOT_STARTED
     )
+    review_status: Mapped[list[TestExecutionReviewDecision]] = mapped_column(
+        ARRAY(Enum(TestExecutionReviewDecision)),
+        default=[],
+    )
```
omar-selo, line 245: Again might be better if it's a set

backend/test_observer/data_access/models.py
```diff
@@ -229,6 +239,11 @@ class TestExecution(Base):
...
+    review_comment: Mapped[str | None] = mapped_column(nullable=False, default="")
```
omar-selo, line 246:
```suggestion
    review_comment: Mapped[str] = mapped_column(default="")
```

backend/tests/controllers/test_executions/test_test_executions.py
```diff
@@ -252,7 +253,7 @@ def test_report_test_execution_data(db_session: Session, test_client: TestClient
     assert test_execution.test_results[1].status == TestResultStatus.SKIPPED
 
 
-def test_updates_test_execution(db_session: Session, test_client: TestClient):
+def _prepare_test_execution_object(db_session: Session) -> TestExecution:
```
omar-selo, line 256: I have no strong opinions on this. It's fine that this is a helper function but it could also be a fixture

omar-selo (changes requested):

backend/test_observer/data_access/models.py
andrejvelichkovski: I think PostgreSQL doesn't have a SET data structure on the database level. Good balance I found is to define it as a set on FastAPI-level and ARRAY on PostgreSQL-level

omar-selo (approved):
LGTM!

andrejvelichkovski:
Forgot to handle the re-run case. Will update the code again.

omar-selo (approved):
Looks good

PR #79, opened 2023-12-11
Fix test execution incorrect status

omar-selo:
Noticed a bug where a test execution that finished was marked as passed even though it failed. This PR fixes that and adds more seed data that include ended tests.

Edit: Also includes some version validations [requested](https://chat.canonical.com/canonical/pl/9gj1catyrj8nxnd9qzwd9nzejo) by @jocave


andrejvelichkovski (dismissed):
Looks good to me!

Thanks for making these changes, I was just curious whether we can make this `seed_data` script always run after the dev server startup, might be nice to always have some data in the DB. No need to do this in this PR, maybe we can do it at some point as a low priority improvement.

omar-selo:
I think it should be doable, using the same method we run migrations after a db is deployed. Not sure if we'll face problems with it though. Also while I don't foresee any uses for an empty database, it could be that we may not want to run seed script every time for some reason

andrejvelichkovski (approved):
Looks good to me!

PR #80, opened 2023-12-11
Frontend charm integration tests fix with a self hosted runner

mz2:
Integration tests for the frontend charm now actually work also on GitHub!

The workflow is stable, with a dozen passes with no intermittent failures over the last few days.

It does however take ~10mins to run and I figured we would therefore not start running it on every commit to every branch (6+ mins of the time is taken by the microk8s setup -- we can drop this in a big way with custom images used for the CI runs when that becomes available). It runs on:
- all branches once a PR review has been posted for the branch for which there is a PR open
- all branches with a PR which touch resources under `./frontend/charm` (the integration tests can capture failures also in resources outside that directory, hence the rule above also).
- on the main branch
- on version tagged commits
- manually via workflow dispatch

The workflow was refactored to reuse two now introduced reusable composite actions in https://github.com/canonical/certification-github-workflows (the [first PR to that repo](https://github.com/canonical/certification-github-workflows/pull/2) should be reviewed and merged first before this goes in).

**NOTE!** When squashing this into main, please edit the commit message heavily down in volume since this contains a few months worth of desperate debugging attempts of these integration tests that continued to work locally just fine but failed to be executed specifically on GitHub up until now. I have simplified the version history with some squashing but since there are several merge commits there I didn't want to do anything more drastic than that.

(Re-posted this to supersede https://github.com/canonical/test_observer/pull/54 -- @plars has gone on much deserved holidays and I had addressed all his valid concerns and questions, with the PR unfortunately left in "changes requested" state)


omar-selo (approved):
LGTM!

PR #81, opened 2023-12-12
GitHub workflow for validating `terraform deploy` + upgrading charms

mz2:
- Introduces a GitHub workflow for validating that `terraform deploy` results in a successful deploy on a fresh microk8s cloud.
  - The new CI checks deploys the plan for the application into a fresh new local microk8s cloud, i.e. it deploys a version of the application presently available on charmhub
  - It then refreshes locally built backend and frontend charms into the running cluster, leading to the upgrade hooks in the charm to be tested as well 
- Fixes some dreadful mutable state driven logical errors in the charm: upon observing the newly introduced CI check fail (1/3 of the application units would launch, with the non-leaders left waiting for the db relation) I proceeded to fix that.
- Removes the `metadata.yml` and moves all the charm metadata to the modern `charmcraft.yml` only format.
- Adds an `online` state check (see https://github.com/canonical/pebble#layer-specification) to the backend charm. 
- Updates charm libraries.
- Limits pointless running of frontend and backend tests on commits that don't change them.


backend/charm/charmcraft.yaml
```diff
@@ -1,8 +1,34 @@
+name: test-observer-api
```
mz2, line 1: I migrated to the new one-yaml format.

backend/charm/src/faults.py
```diff
@@ -0,0 +1,18 @@
+class UnitReadinessFault:
+    """A ReadinessFault describes the reason for a unit not to be ready."""
+
+    def __init__(self, reason: str, parent=None):
```
mz2, line 4: I experienced a Python noob issue here that I worked around by ignoring type hinting the `parent` here: I tried adding a type hint `parent=UnitReadinessFault | None` here but VS Code with mypy running didn't appreciate that. I also tried stringifying it, which made VS Code happy but caused a runtime issue with `"str" | None` being taken at charm runtime to be the type.

backend/charm/src/charm.py
```diff
@@ -61,40 +59,54 @@ def _setup_nginx(self):
...
 
-    def _migrate_database(self):
-        # only leader runs database migrations
+    def _attempt_database_migration(self) -> bool | UnitReadinessFault:
+        """Return true if migrations were attempted, false if not attempted (if unit is not the leader).
+
+        If unit is not ready for migrations, return UnitReadinessFault.
+        """
         if not self.unit.is_leader():
-            raise SystemExit(0)
```
mz2, line 69: I moved away from this needless `SystemExit(0)` business everywhere I saw.

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "python.analysis.extraPaths": [
+        "./backend/charm/lib"
+    ],
+    "cmake.sourceDirectory": "/home/mz2/Developer/test_observer/frontend/linux"
```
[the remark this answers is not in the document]
mz2: Yes, this was intentional since VS Code looked like it was now doing relative path updates. However, what I committed was the wrong state of the file :facepalm:

backend/charm/src/charm.py
```diff
@@ -61,40 +59,54 @@ def _setup_nginx(self):
...
-        self._migrate_database()
+        if not self._attempt_database_migration():
+            self.unit.status = WaitingStatus("Waiting for database relation")
+
         self._update_layer_and_restart(None)
 
     def _on_database_relation_broken(self, event):
-        self.unit.status = WaitingStatus("Waiting for database relation")
+        self.unit.status = WaitingStatus("Waiting for database relation after breaking relation")
         raise SystemExit(0)
```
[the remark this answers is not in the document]
mz2: Neither should be done, it should just return.

mz2, line 110:
```suggestion
```

PR #82, opened 2023-12-12
Add get test results endpoint

omar-selo:
Adds an endpoint to fetch individual test results. This is needed for [RTW-202](https://warthogs.atlassian.net/browse/RTW-202)

[RTW-202]: https://warthogs.atlassian.net/browse/RTW-202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/test_observer/controllers/test_executions/models.py
+++ b/backend/test_observer/controllers/test_executions/models.py
@@ -22,14 +22,20 @@ from enum import Enum
 from typing import Annotated
 
 from pydantic import (
+    AliasPath,
     BaseModel,
+    Field,
     HttpUrl,
     field_serializer,
     field_validator,
     model_validator,
 )
 
-from test_observer.data_access.models_enums import FamilyName, TestExecutionStatus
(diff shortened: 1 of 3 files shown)


andrejvelichkovski:
Wouldn't it be easier and more performance-efficient to include those in the `artefact/<ID>/builds` endpoint, as we already have Test Executions there

omar-selo:
Issue is that there are lots of test results and their io log can be hefty. So if we pull everything it might be too big of a request. I was that this new endpoint would only be called whenever a user clicks on the drop down icon of a test execution. Cause in most cases the user might actually not care about looking at individual test results. Just the test executions. If they do choose to see results then we'll fetch them

andrejvelichkovski (approved):
Thanks for clarifying on the reasoning. I agree some tests have a lot of data associated with them, so loading all of them in the same time might actually hurt performance.

The changes look good to me!

PR #83, opened 2023-12-12
Adds manual workflow dispatches

mz2:


omar-selo (commented):
Issue with this is that when say you publish api it publish oci image and charm. The oci image is tagged with branch name. And the charm pulls oci image from main so it doesn't get the latest image

mz2:
Boo

PR #84, opened 2023-12-13
Show individual test results on frontend

omar-selo:
Related to [RTW-202](https://warthogs.atlassian.net/browse/RTW-202). This PR displays the names and statuses of individual test results. I've also updated the seed script to put real test results.

Notes:
1. I'll add more details on test result in future PR
2. Sadly right now test results are all fetched as soon as you click on an artefact. Need to figure out how to only fetch them when expanding an environment

See below screenshot:
image

[RTW-202]: https://warthogs.atlassian.net/browse/RTW-202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/scripts/seed_data.py
```diff
@@ -165,51 +167,89 @@
...
+3
+ Container 4506-test-1  Running
+ Container 4506-test-1  Stopping
+ Container 4506-test-1  Stopping
+ Container 4506-test-1  Stopped
+ Container 4506-test-1  Removing
+ Container 4506-test-1  Removed
+ Network 4506_default  Removing
+ Network 4506_default  Removed
+""",
```
andrejvelichkovski, line 206: Nadzeya pointed out the textwrap module that has useful [dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) function in another PR a few days ago. Maybe we can apply it here to make the text indent in the same way as the object arguments.

frontend/lib/ui/dashboard/artefact_dialog.dart
```diff
@@ -302,7 +304,73 @@ class _TestExecutionView extends StatelessWidget {
...
+                ),
+              )
+              .toList(),
+        ),
+      ),
+    );
+  }
+}
+
+class _TestResultsFilter extends ConsumerWidget {
```
andrejvelichkovski, line 324: Not sure if Flutter has a rule around how Widgets should be organised in different files, but I see this one has grown to over 300 lines. Could we split this method in a new file, maybe `test_case_results.dart` or even the whole TestExecution view in `test_executions.dart`, might be even more helpful when we start adding the signoff stuff. What do you think about this?

andrejvelichkovski (commented):
The logic behind the change looks good overall to me! I left two small comments in the code, but my bigger question is whether we should present the IOLog and the test comments on the frontend.

omar-selo:
> is whether we should present the IOLog and the test comments on the frontend.

I was planning to add them in a later PR. Or do you mean whether we should do so at all?

andrejvelichkovski:
I think we should add them, but doing it in a later PR makes sense, easier to review :)

andrejvelichkovski (dismissed):
Looks good to me!

backend/scripts/seed_data.py
omar-selo: Oh nice find, will use it

frontend/lib/ui/dashboard/artefact_dialog.dart
omar-selo: Completely agree, I've been neglecting this problem for too long. Will separate them into files

nadzyah:
Is it possible to represent the individual test names as a bullet list? What do you think about it?

backend/scripts/seed_data.py
```diff
@@ -165,51 +169,91 @@
...
             C3TestResult(
-                id=3,
-                name="test3",
+                id=2,
+                name="after-suspend-audio/alsa-loopback-automated",
                 status=C3TestResultStatus.SKIP,
-                category="",
-                comment="",
+                category="Audio tests",
+                comment="""job cannot be started: resource expression "manifest.has_audio_loopback_connector == 'True'" evaluates to false, required dependency 'com.canonical.certification::audio/alsa-loopback-automated' has failed, required dependency 'com.canonical.certification::audio/detect-capture-devices' has failed, required dependency 'com.canonical.certification::audio/detect-playback-devices' has failed, required dependency 'com.canonical.certification::suspend/suspend_advanced_auto' has failed""",
```
nadzyah, line 217: are you sure you don't want to split this string?

omar-selo: Tried to, including the other long lines, but it got uglier and more complicated :/

andrejvelichkovski (approved):
Looks really nice having all these smaller files instead of a large one. Thanks for fixing this!

backend/scripts/seed_data.py
```diff
@@ -165,51 +169,91 @@
...
+                    Container 4506-test-3  Created
+                    Container 4506-test-2  Created
+                    Container 4506-test-1  Recreated
+                    Container 4506-test-1  Starting
+                    Container 4506-test-1  Started
+                    Container 4506-test-2  Starting
+                    Container 4506-test-2  Started
+                    Container 4506-test-3  Starting
+                    Container 4506-test-3  Started
+                    3
```
nadzyah, line 200: I'm wondering if it's fine to have this random number in this list

omar-selo: This is just a string that isn't parsed so it doesn't break anything. But I do agree with you this random number is weird. So I went back to the [source](https://certification.canonical.com/hardware/202011-28409/submission/345403/test/90212/result/37922493/) to make sure I copied it correctly. And sure enough the 3 exists :sweat_smile:

nadzyah (approved):

PR #85, opened 2023-12-14
Fix flaky test because of unordered sets

andrejvelichkovski:
This PR fixes a flaky test that sometimes fails because [sets](https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset) in Python are unordered data structures. Because of this, the order they might return the results is unpredictable, and depends on the current hash value of the objects.

To account for this, we instead verify the length of the set and that the result expected indeed are present in the set.

### DIFF
--- a/backend/tests/controllers/test_executions/test_test_executions.py
+++ b/backend/tests/controllers/test_executions/test_test_executions.py
@@ -273,30 +273,27 @@ def test_report_test_execution_data(db_session: Session, test_client: TestClient
 def test_updates_test_execution(
     db_session: Session, test_client: TestClient, test_execution: TestExecution
 ):
+    new_review_decision = set(
+        [
+            TestExecutionReviewDecision.APPROVED_FAULTY_HARDWARE.name,
+            TestExecutionReviewDecision.APPROVED_INCONSISTENT_TEST.name,
+        ]
+    )
     test_client.patch(
         f"/v1/test-executions/{test_execution.id}",
         json={
             "ci_link": "http://ci_link/",
             "c3_link": "http://c3_link/",
             "status": TestExecutionStatus.PASSED.name,
-            "review_decision": [
-                TestExecutionReviewDecision.APPROVED_FAULTY_HARDWARE.name,
-                TestExecutionReviewDecision.APPROVED_INCONSISTENT_TEST.name,
-            ],
+            "review_decision": list(new_review_decision),
             "review_comment": "Tests fail because of broken keyboard",
         },
     )
-
     db_session.refresh(test_execution)
     assert test_execution.ci_link == "http://ci_link/"
     assert test_execution.c3_link == "http://c3_link/"
     assert test_execution.status == TestExecutionStatus.PASSED
-    assert len(test_execution.review_decision) == 2
-    for review_decision in [
-        TestExecutionReviewDecision.APPROVED_FAULTY_HARDWARE.name,
-        TestExecutionReviewDecision.APPROVED_INCONSISTENT_TEST.name,
-    ]:
-        assert review_decision in test_execution.review_decision
+    assert set(test_execution.review_decision) == new_review_decision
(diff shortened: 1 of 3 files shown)


backend/tests/controllers/test_executions/test_test_executions.py
```diff
@@ -291,10 +291,12 @@ def test_updates_test_execution(
...
     assert test_execution.c3_link == "http://c3_link/"
     assert test_execution.status == TestExecutionStatus.PASSED
-    assert test_execution.review_decision == [
+    assert len(test_execution.review_decision) == 2
+    for review_decision in [
         TestExecutionReviewDecision.APPROVED_FAULTY_HARDWARE.name,
         TestExecutionReviewDecision.APPROVED_INCONSISTENT_TEST.name,
-    ]
+    ]:
+        assert review_decision in test_execution.review_decision
```
omar-selo, line 299: while this is fine, a more pythonic way would be to just compare sets. If you define a set say `review_decision` then you can just compare it via `review_decision == test_execution.review_decision`. If you define it as a list you could just call `set()` on it

andrejvelichkovski: I updated this. However, I came across an interesting problem I can't understand. The `test_execution.review_decision` still seems to be of a type <list>. However, when I tested this endpoint, it indeed showed `set` behaviour. Even if I passed the same argument twice or more, it still compressed it into one. I'm not sure what might be the cause of this.

backend/tests/controllers/test_executions/test_test_executions.py
```diff
@@ -273,28 +273,25 @@ def test_report_test_execution_data(db_session: Session, test_client: TestClient
...
-
     db_session.refresh(test_execution)
     assert test_execution.ci_link == "http://ci_link/"
     assert test_execution.c3_link == "http://c3_link/"
     assert test_execution.status == TestExecutionStatus.PASSED
-    assert test_execution.review_decision == [
-        TestExecutionReviewDecision.APPROVED_FAULTY_HARDWARE.name,
-        TestExecutionReviewDecision.APPROVED_INCONSISTENT_TEST.name,
-    ]
+    assert set(test_execution.review_decision) == set(new_review_decision)
```
omar-selo, line 294: hmm so runtime type of `test_execution.review_decision` is not a set? even though it's mapped as set?

andrejvelichkovski: Yeah, I guess the set behaviour we get is because it is defined as set in `TestExecutionsPatchRequest`

backend/tests/controllers/test_executions/test_test_executions.py
omar-selo: I guess it's fine if the endpoint just converted the request to a set

omar-selo (dismissed):
Looks good

backend/tests/controllers/test_executions/test_test_executions.py
andrejvelichkovski: Sure, I now updated it to match this. `set` is not a native datatype in PostgreSQL, so the list data type on this level matches this as well. I also think keeping this data type on the request-level makes sense.

andrejvelichkovski: It seems like we must keep this set data type on this level as well, otherwise mypy complains that we are trying to assign set to a list in the API logic. Another alternative would be to define everything as a list.

omar-selo: I'd go with either everything is a list or we convert the request set to a list when passing to model

omar-selo: set on model is currently sort of a lie

andrejvelichkovski: True, I will keep the set on request-level but convert to list whenever assigning to the DB model

omar-selo (approved):

PR #90, opened 2023-12-15
Artefact signoff frontend

omar-selo:
Mostly resolves [RTW-108](https://warthogs.atlassian.net/browse/RTW-108).

Includes the following changes:
- Refactor dashboard_body.dart file by removing individual widgets into their own files
- Display artefact status on the artefact card
- In artefact dialog adds a button that changes artefact status (next to artefact name)
- Sort artefacts fetched from backend by name then created_at. This is to avoid the order changing whenever someone signs off an artefact (as looks like default sorting is via updated at)
- Make backend artefact patch endpoint return the updated artefact. This is so that the frontend updates the stored value of the artefact without making another request to fetch it

Current issues:
- The main issue is that when you update an artefact's status in the artefact dialog it doesn't update automatically in the dashboard. You have to refresh the page to see the updates. I know why this is the case but fixing it is not simple. At least it would require time. So we can leave it to another PR.

Note:
- Ignore the slowness in the video as my machine is struggling

[Screencast from 15-12-23 16:57:36.webm](https://github.com/canonical/test_observer/assets/4087200/84302eb8-9b23-4be2-8a11-5a56da8c266b)

[RTW-108]: https://warthogs.atlassian.net/browse/RTW-108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andrejvelichkovski (approved):
Looks good to me! We can focus on resolving the refresh issue later.

PR #92, opened 2024-01-02
Add api requests retries on frontend

omar-selo:
Occasionally, api requests sent by frontend have transient failures. This can result in errors as shown below:
image

To avoid or minimize such errors this PR adds retries.

### DIFF
--- a/frontend/lib/providers/dio.dart
+++ b/frontend/lib/providers/dio.dart
@@ -4,6 +4,7 @@ import 'dart:html';
 import 'dart:js_util' as js_util;
 
 import 'package:dio/dio.dart';
+import 'package:dio_smart_retry/dio_smart_retry.dart';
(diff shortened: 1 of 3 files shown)


andrejvelichkovski (approved):
I think most of these errors will be gone when we fix the two broken units, but I think it is still a good practice to have some resilience to network problems. LGTM!

Btw, I was just curious if we need to setup some parameters, like max number of retries, or these are automatically setup by the retry library?

omar-selo:
> I was just curious if we need to setup some parameters

These are set to some defaults. Namely, 3 retries, with the duration being 1 second, 3 seconds, and 5 seconds.

andrejvelichkovski:
> These are set to some defaults. Namely, 3 retries, with the duration being 1 second, 3 seconds, and 5 seconds.

That makes sense, thanks for clarifying it!

PR #93, opened 2024-01-02
Implement frontend for test execution review

andrejvelichkovski:
This PR resolves RTW-204 and implements the frontend functionality for test execution review.

# Implementation Changes

As part of this change, a few things have been updated:
* New popover for the review process was added
* Button to trigger the popover
* Riverpod provider ArtefactBuilds was updated to also include functionality for patching the test execution

# Screenshots
Screenshot from 2024-01-03 09-17-27
Screenshot from 2024-01-03 09-17-34
Screenshot from 2024-01-03 09-17-48

# Design changes

I have changed the design slightly, as I couldn't find a way to use `YaruPopupMenuButton` which is specified in the design to be a standard button like TextButton.

# Unresolved issues

When updating an existing Test Execution Review, I managed to load the selected decisions. However, I couldn't find a way to load the review comment into the `TextField` widget.


frontend/lib/ui/artefact_dialog/test_execution_review.dart
```diff
@@ -0,0 +1,188 @@
...
+              hintText: 'Insert review comment',
+            ),
+            onChanged: (value) {
+              setState(() {
+                reviewComment = value;
+              });
+            },
+          ),
+          Padding(
+            padding: const EdgeInsets.only(top: 8.0),
```
omar-selo, line 172: typically spacing between widgets is done using `SizedBox` rather than padding

omar-selo (changes requested):
Thanks for the changes. Most are small points but one issue is about passing `artefactId`

frontend/lib/providers/artefact_builds.dart
```diff
@@ -1,20 +1,40 @@
...
+    await dio.patch(
+      '/v1/test-executions/$testExecutionId',
+      data: {
+        'review_decision': reviewDecision.map((e) => e.toJson()).toList(),
+        'review_comment': reviewComment,
+      },
+    );
+
+    final updatedState = await build(artefactId);
+    state = AsyncData(updatedState);
```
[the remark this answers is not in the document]
andrejvelichkovski: I was also thinking about this, but the issue with the first option was that the PATCH endpoint modified the TestExecution object, not the ArtefactBuild object, and the riverpod provider is for the ArtefactBuilds. 

I guess one option would be to have this endpoint return the whole ArtefactBuild object, which is a bit counter-intuitive as we are sending PATCH request to a TestExecution object.

frontend/lib/ui/artefact_dialog/test_execution_review.dart
```diff
@@ -0,0 +1,188 @@
...
+  bool _enableApprovedOptions() {
+    return !(reviewDecisions[TestExecutionReviewDecision.rejected] ?? false);
+  }
+
+  bool _enableRejectedOptions() {
+    return !TestExecutionReviewDecision.values.any(
+      (element) => (element != TestExecutionReviewDecision.rejected &&
+          (reviewDecisions[element] ?? false)),
+    );
+  }
```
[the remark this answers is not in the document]
andrejvelichkovski: I'm not sure if you mean getter on the TestExecutionReviewDecision enum. In that case, I'm not sure how we can do it, as these are stateful (they depend on the list of selected decisions).

frontend/lib/providers/artefact_builds.dart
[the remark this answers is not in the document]
omar-selo: True, I hadn't taken that into consideration. Then I guess options 2 or 3 would work. Where for 3 we would just replace the child test execution object in the artefact build

frontend/lib/ui/artefact_dialog/test_execution_review.dart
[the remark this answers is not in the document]
omar-selo: Oh I wasn't very clear. I mean class getters https://dart.dev/language/methods#getters-and-setters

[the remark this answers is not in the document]
andrejvelichkovski: Thanks! I didn't know about getters/setters, but will update the code.

frontend/lib/ui/artefact_dialog/artefact_build_expandable.dart
```diff
@@ -7,9 +7,14 @@ import '../spacing.dart';
 import 'test_execution_expandable.dart';
 
 class ArtefactBuildExpandable extends StatelessWidget {
-  const ArtefactBuildExpandable({super.key, required this.artefactBuild});
+  const ArtefactBuildExpandable({
+    super.key,
+    required this.artefactBuild,
+    required this.artefactId,
+  });
```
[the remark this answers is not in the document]
andrejvelichkovski: I agree passing the prop is not perfect, but I wasn't sure how to do it otherwise.

We don't need the artefactId to send this request, only the test execution id, but we need the artefactID to get the ArtefactBuild object from riverpod.

I guess the best approach would be to have TestExecutionProvier which would not send requests to the backend but rather take the data from the ArtefactBuildsProvider, similarly to how we have the StageArtefactsProvider take the data from FamilyArtefactsProvider.

However, my only concern is how we would update the ArtefactBuildProvider state in that case, as it might include some incosistent data with the TestExecutionProvider, in case we change something. I guess in that case we need to make sure to never use the .testExecution object from ArtefactBuildProvider, but to always use the TestExecutionProvider, which makes sense.

I'm not sure about the flutter data package, as I don't have a strong understanding of how Flutter works yet. I will try to read more around it.

How do you suggest we go about this?

frontend/lib/ui/artefact_dialog/test_execution_review.dart
```diff
@@ -0,0 +1,195 @@
...
+      onTap: () {
+        showPopover(
+          context: context,
+          bodyBuilder: (context) => TestExecutionPopOver(
+            testExecution: testExecution,
+            artefactId: artefactId,
+          ),
+          direction: PopoverDirection.bottom,
+          width: 500,
+          height: 500,
```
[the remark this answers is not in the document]
andrejvelichkovski: Yes, the popover becomes scrollable

Screenshot from 2024-01-03 12-21-30

frontend/pubspec.yaml
```diff
@@ -15,6 +15,7 @@ dependencies:
   go_router: ^11.1.4
   intersperse: ^2.0.0
   json_annotation: ^4.8.1
+  popover: 0.2.8
```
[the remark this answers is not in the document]
andrejvelichkovski: It kept downloading version 0.2.9 for me locally, which wasn't working with the Git checker. So I fixed it to 0.2.8. I will check the automated checks to see if they fail again

frontend/lib/providers/artefact_builds.dart
```diff
@@ -1,20 +1,40 @@
...
+    String reviewComment,
+    List<TestExecutionReviewDecision> reviewDecision,
+  ) async {
+    final dio = ref.watch(dioProvider);
+
+    await dio.patch(
+      '/v1/test-executions/$testExecutionId',
+      data: {
+        'review_decision': reviewDecision.map((e) => e.toJson()).toList(),
+        'review_comment': reviewComment,
```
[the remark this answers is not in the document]
andrejvelichkovski: I have refactored to have these value as a constant and generate inside the class

frontend/lib/providers/artefact_builds.dart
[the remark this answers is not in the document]
andrejvelichkovski: I was thinking it would be cleaner to go with the second approach, although it involves an extra network request to rebuild the state, but it makes sure we always put the correct data in the state.

frontend/lib/providers/artefact_builds.dart
```diff
@@ -19,22 +19,47 @@ class ArtefactBuilds extends _$ArtefactBuilds {
...
+            artefactBuild.testExecutions.map(
+              (e) {
+                if (e.id == testExecutionId) {
+                  return TestExecution.fromJson(responseData);
+                }
+                return e;
+              },
+            ),
+          ),
+        ),
```
omar-selo, line 42: I don't think that this code is completely correct because at doesn't make copies of other test executions and doesn't copy environments. And state [should be immutable](https://riverpod.dev/docs/concepts/why_immutability#using-immutable-state). I recommend using `copyWith` for artefact builds and test executions and possibly adding a helper function.

frontend/lib/ui/artefact_dialog/test_execution_pop_over.dart
```diff
@@ -0,0 +1,112 @@
...
+                      ? (bool? value) {
+                          setState(() {
+                            if (reviewDecisions.contains(e)) {
+                              reviewDecisions.remove(e);
+                            } else {
+                              reviewDecisions.add(e);
+                            }
+                          });
+                        }
+                      : null,
```
omar-selo, line 77: can you pull `onChanged` as a separate helper method/function as it's a bit hard to follow this way

frontend/lib/ui/artefact_dialog/test_execution_review.dart
```diff
@@ -40,6 +37,7 @@ class TestExecutionReviewButton extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final artefactId = AppRoutes.artefactIdFromContext(context);
```
omar-selo, line 40: given that this widget doesn't use `artefactId` perhaps `TestExecutionPopOver` should be the one to get it from context?

omar-selo (changes requested):
Looks much better, just a few more comments

frontend/lib/ui/artefact_dialog/test_execution_review.dart
andrejvelichkovski: Actually, I tried this initially and it didn't work, I think it is because the Popover is a separate layer in the render. After some searching, I found out about this [open issue in Flutter](https://github.com/flutter/flutter/issues/131154) which seems to be going resolved in a [separate issue](https://github.com/flutter/flutter/issues/129833) but doesn't seem to be closed yet.

backend/test_observer/data_access/models.py
```diff
@@ -166,7 +166,9 @@ class ArtefactBuild(Base):
         back_populates="builds", foreign_keys=[artefact_id]
     )
     test_executions: Mapped[list["TestExecution"]] = relationship(
-        back_populates="artefact_build", cascade="all, delete"
+        back_populates="artefact_build",
+        cascade="all, delete",
+        order_by="TestExecution.environment_id",
```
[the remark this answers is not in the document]
andrejvelichkovski: Good point, I have updated to order it at the controller level

frontend/lib/providers/artefact_builds.dart
andrejvelichkovski: I see, this code is indeed not really good idea, but I think using copyWith wont be much simpler, as we again need to rebuild at least one artefact build. It seems like a better option is to make that additional network request after all..

omar-selo: adding a network request feels like choosing a worst solution for the sake of better code. We don't need to rebuild one artefact build. Once we copyWith everything, we just modify one artefact build

omar-selo: Oh ignore my point about modifying one artefact build as it's a freezed object so it can't be modified

andrejvelichkovski: I agree with your points, I think my implementation should be updating the state correctly now.

frontend/lib/ui/artefact_dialog/test_execution_pop_over.dart
andrejvelichkovski: I now extracted this in a separate function.

frontend/lib/ui/artefact_dialog/artefact_dialog.dart
```diff
@@ -16,47 +17,59 @@ class ArtefactDialog extends ConsumerWidget {
...
+                        crossAxisAlignment: CrossAxisAlignment.start,
+                        children: [
+                          ArtefactDialogHeader(artefact: artefact),
+                          const SizedBox(height: Spacing.level4),
+                          ArtefactDialogInfoSection(artefact: artefact),
+                          const SizedBox(height: Spacing.level4),
+                          Expanded(
+                            child: ArtefactDialogBody(artefact: artefact),
+                          ),
+                        ],
```
andrejvelichkovski, line 63: This is the change to fix the async error.

andrejvelichkovski:
> It looks like the c3 link is still there in the code, but I don't see it in the screenshots you posted. Is that just something off with the screenshot, or was there some UI change that somehow removed the c3 link here?

The C3 Link button is conditionally rendered, it just happened that none of these test executions have it set up in the database, so it is not shown on the frontend either.

frontend/lib/ui/artefact_dialog/artefact_dialog.dart
omar-selo: Thanks for fixing my error

PR #96, opened 2024-01-05
Add condition to check if C3 Link is found

andrejvelichkovski:
For older artefacts, before we added a check that the C3 link must be a valid URL, Jenkins was sending the text "Failed to find C3 submission link" as C3 link. So now, some snaps (one example: https://test-observer.canonical.com/#/snaps/11035) show the C3 link button, but it doesn't point anywhere.

To fix this, we add the condition that checks for this case.


omar-selo (changes requested):
A better solution to this problem would be to just empty those strings in the database. There's reason for our code to handle this special case given that it doesn't need to be there

andrejvelichkovski:
I agree with you suggestion, I will prepare a query to execute in the database and verify it before doing so. Closing this PR as it is not needed.

PR #97, opened 2024-01-05
Implement functionality for getting historic test results

andrejvelichkovski:
This change implements the backend for RTW-120. It extends the response for the `/test-executions/<ID>/test-results` with an additional field `historical_results` for each test result field.

# Implementation

As part of this implemention there are a few modifications on the backend:
* new repository method is implemented `get_historic_test_results` together with a helper method `_get_historic_test_executions_ids` 
* extended the `/test-executions/<ID>/test-results` endpoint to also include this data
* extended both the TestResult and TestResultDTO models to have the `historic_results` fields

# Testing

## Manual testing

I adapted the `seed_data.py` to initialise many test executions in the same environment and triggered the GET endpoint `/test-executions/<ID>/test-results`. This is how a [sample response](https://pastebin.canonical.com/p/k4TY9syj8J/) to one of the calls.

## Automated testing

Existing test for this endpoint was modified to include a test for this functionality.


andrejvelichkovski:
This PR needs changes, as after some discussions, we realised we want to track the version number for each test result status. Converting it to draft for now.

backend/test_observer/data_access/helpers.py
```diff
@@ -0,0 +1,30 @@
+from sqlalchemy.orm import Session
+
+from test_observer.data_access.models import Artefact, ArtefactBuild, TestExecution
+
+HISTORIC_TEST_RESULT_COUNT = 10
```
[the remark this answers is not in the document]
andrejvelichkovski: Sure thing, will address this.

backend/test_observer/data_access/models.py
```diff
@@ -296,6 +295,8 @@ class TestResult(Base):
     )
     test_case: Mapped["TestCase"] = relationship()
 
+    historic_results: list = []
+
```
omar-selo, line 299: Perhaps this property is not needed?

backend/test_observer/data_access/repository.py
```diff
@@ -130,6 +139,35 @@ def get_artefacts_by_family(
...
+            )
+        )
+        .filter(
+            Artefact.name == current_artefact.name,
+            Artefact.store == current_artefact.store,
+            Artefact.repo == current_artefact.repo,
+            Artefact.series == current_artefact.series,
+            TestExecution.environment_id == test_execution.environment_id,
+            TestExecution.id < test_execution.id,
+        )
```
omar-selo, line 163: perhaps you meant to use `Artefact.track == current_artefact.track` instead of `Artefact.store == current_artefact.store`

backend/test_observer/data_access/repository.py
```diff
@@ -130,6 +139,35 @@ def get_artefacts_by_family(
     return query.all()
 
 
+def get_historic_test_executions(
+    session: Session,
+    test_execution: TestExecution,
+) -> list[TestExecution]:
```
omar-selo, line 145: In my mind a repository is more of a simple wrapper that provides some helper methods to access the database. Perhaps business logic should not reside in it. I would move this function to [here](https://github.com/canonical/test_observer/blob/main/backend/test_observer/controllers/test_executions/logic.py)

backend/test_observer/controllers/test_executions/models.py
```diff
@@ -124,10 +124,16 @@ def validate_review_decision(
...
+
+
 class TestResultDTO(BaseModel):
     id: int
     name: str = Field(validation_alias=AliasPath("test_case", "name"))
     category: str = Field(validation_alias=AliasPath("test_case", "category"))
     status: TestResultStatus
     comment: str
     io_log: str
+    historic_results: list[HistoricTestResult] = []
```
omar-selo, line 139: I would suggest adding a [description](https://docs.pydantic.dev/latest/concepts/json_schema/) to historic_results specifying the order (desc or asc). This would appear at the api docs

backend/test_observer/controllers/test_executions/test_executions.py
```diff
@@ -190,4 +196,14 @@ def get_test_results(id: int, db: Session = Depends(get_db)):
...
 
+    historic_test_executions = get_historic_test_executions(db, test_execution)
+    historic_test_result_mapping = get_historic_test_result_mapping(
+        historic_test_executions
+    )
+
+    for test_result in test_execution.test_results:
+        test_result.historic_results = historic_test_result_mapping[
+            test_result.test_case_id
+        ]
```
omar-selo, line 207: what would happen if there is a new test case not tested before?

backend/test_observer/data_access/repository.py
```diff
@@ -130,6 +139,35 @@ def get_artefacts_by_family(
...
+            Artefact.repo == current_artefact.repo,
+            Artefact.series == current_artefact.series,
+            TestExecution.environment_id == test_execution.environment_id,
+            TestExecution.id < test_execution.id,
+        )
+        .options(joinedload(TestExecution.test_results, innerjoin=True))
+        .order_by(TestExecution.created_at.desc())
+        .limit(HISTORIC_TEST_RESULT_COUNT)
+        .all()
+    )
```
omar-selo, line 168: One suggestion expand on this SQL query to retrieve the test results grouped by test case and sorted already. This is likely more efficient

omar-selo (changes requested):
Looks good, just a few points

backend/test_observer/data_access/models.py
andrejvelichkovski: This property is actually needed. The reason for this seems to be that the conversion from TestResult -> TestResultDTO happens after we return the response for this endpoint. Because of this, we don't have defined the historic_results property for TestResults and mypy complains about it.

backend/test_observer/data_access/repository.py
andrejvelichkovski: Thanks for catching this, I forgot track is what we use to identify the artefact.

backend/test_observer/controllers/test_executions/models.py
andrejvelichkovski: I have now added description to this field

backend/test_observer/controllers/test_executions/test_executions.py
andrejvelichkovski: Thanks for catching this, it would throw a KeyError. I have now fixed this.

backend/test_observer/data_access/repository.py
andrejvelichkovski: I have now rewritten this. The only issue I couldn't resolve is how to generate a tuple of (version, status) on SQL-level, so I return two separate arrays, one for status, one for version and parse them into HistoricTestResult object in Python. Do you have any suggestions on how to do this better?

backend/test_observer/data_access/models.py
omar-selo: Oh I see. Then there should be a way to pass it directly to the pydantic model or perhaps a serialization only field. Since this is not an orm field

andrejvelichkovski: I was thinking only Mapped fields are considered ORM fields, and the others are serialization-level fields. But, I will look at how we can convert the object to DTO object at request-level.

andrejvelichkovski: I resolved the issue by converting the test results to TestResultDTO when parsing, and then added the historic results to this field.

backend/test_observer/data_access/repository.py
omar-selo: After investigating this further I found that this query is quite complex. Looking at your latest code I think (not sure) that it has issues. Perhaps we should simplify this into separate problems and it should be ok to make use of python (I take back doing everything in sql land, at least group by is much easier on python land). It would be great to test all scenarios. But that will be challenging if they are all lumped into one function. So having smaller functions each with one purpose (even if each function is a sql subquery) would be much easier to test.

In high level we need the following:
1. Latest artefact builds per artefact and architecture. This would be a sql stmt and would be tested alone. We may already have code for this. Although I recall it uses group by. According to [this stackoverflow answer](https://stackoverflow.com/a/1313293), it's much faster to use self join over group by.
2. Latest test executions per artefact build and environment. Again this would be a sql stmt and would be tested alone. The solution will probably be very similar to the one above.
3. One big sql join that brings artefacts, latest artefact builds, latest test executions, environments, test results and test cases together. We could also sort by artefact id here.
4. One big python group by. Python's group by would create an iterable of grouped items which is what we need. This group by would need to use artefact id, artefact track, artefact series, artefact repo (not sure if group by can handle nulls the way we want, we may have to separate them into two depending on family), environment, test case.

Might have missed something, and there might be further room for improvement. But most importantly we need to test this, and testing individual properly parts is much easier than testing the whole properly

omar-selo: You're absolutely right, maybe I worded it wrong. By artefact builds per architecture and artefact, I meant that the architecure is taken into consideration as to avoid different merging different architectures. But we still need to only get the latest build for each architecture. Because each architecure could have multiple builds. The same is for test executions, although now that I think more about it, this may not be necessary for test executions. Since we have a unique constraint on artefact build id and environment id. This means we can't have two test executions for the same environment and build (it just overrides the previous one). Therefore it's probably safe to skip the point about latest test executions

andrejvelichkovski: I see, you are right that if we have a list of artefact builds, for each of them there will be at most 1 Test Execution we need to consider.

So I guess your suggestion is to find the historic test results in a top-to-bottom style:
1. Find all matching artefact builds, given Artefact identifiers (artefact name, track, series, repo) and Architecture
2. With the matched Artefact Builds from (1), we find the related TestExecutions and filter based on Environment ID
3. Find all the test results associated with the latest 10 test executions from (2)
4. Parse and group them on Python level

omar-selo: Sounds about right, hopefully...

Actually, this begs the question. If we don't allow multiple test executions of the same environment and build, why do we allow multiple builds of the same architecture and artefacts.

andrejvelichkovski: Hmm, I think according to the way we start test executions, there shouldn't be duplicated artefact builds. If I'm not wrong the revision count is sort-of identifier after the artefact_id. Although we don't have a constraint for this on the db-level. I wonder if we have a case of duplicated artefact builds with the same revision number.

omar-selo: Hmmm, you're right, we actually have a [unique index](https://github.com/canonical/test_observer/blob/main/backend/test_observer/data_access/models.py#L172-L190) on artefact_id, architecture and revision (if it existed). So that would mean we can only have one artefact build per artefact and architecture combination. That simplifies things, even in other parts of the code (we were trying to only get latest builds :sweat_smile:)

omar-selo: Oh wait nvm I mean different revisions should be merged or rather only the last revision should be taken into consideration

andrejvelichkovski: @omar-selo could you please take a look at whether my implementation now fits into what you were suggesting? I can add more tests for different edge-cases if needed.

backend/tests/controllers/test_executions/test_logic.py
```diff
@@ -0,0 +1,100 @@
...
+        session=db_session,
+        artefact=test_execution_one.artefact_build.artefact,
+        architecture=test_execution_one.artefact_build.architecture,
+    )
+
+    assert artefact_build_ids == [
+        test_execution_three.artefact_build_id,
+        test_execution_two.artefact_build_id,
+        test_execution_one.artefact_build_id,
+    ]
```
nadzyah, line 92: sorry if it's a stupid question, but does the order matter here? I mean, if it's not important, probably we just should compare sets here?

backend/tests/controllers/test_executions/test_logic.py
```diff
@@ -0,0 +1,100 @@
...
+        artefact=artefact_one,
+        architecture=artefact_build_two.architecture,
+    )
+
+    assert artefact_build_ids == [artefact_build_two.id, artefact_build_one.id]
+
+
+def _get_test_execution(
+    db_session: Session, environment: Environment, ci_link: str
+) -> TestExecution:
```
nadzyah, line 45: what do you think about placing this function before other functions so we don't mix tests and helpers?

nadzyah (commented):
Just some minor comments

backend/tests/controllers/test_executions/test_logic.py
andrejvelichkovski: Yes, the order actually matters here, since we want to get the latest only.

backend/tests/controllers/test_executions/test_logic.py
andrejvelichkovski: Good idea! I have now moved it on the top of the tests

backend/test_observer/controllers/test_executions/models.py
```diff
@@ -130,10 +131,19 @@ class HistoricTestResult(BaseModel):
...
     name: str = Field(validation_alias=AliasPath("test_case", "name"))
     category: str = Field(validation_alias=AliasPath("test_case", "category"))
     status: TestResultStatus
     comment: str
     io_log: str
-    historic_results: list[HistoricTestResult] = []
+    historic_results: list[HistoricTestResult] = Field(
+        default=[],
+        description=(
+            "The last 10 test results matched with the current test execution. "
```
omar-selo, line 145: very small point here, '10' could be replaced with `HISTORIC_TEST_RESULT_COUNT`

backend/test_observer/controllers/test_executions/logic.py
```diff
@@ -49,3 +58,124 @@ def parse_c3_test_result_status(status: C3TestResultStatus) -> TestResultStatus:
...
             return TestResultStatus.SKIPPED
+
+
+def get_historic_artefact_builds_query(
+    session: Session,
+    artefact: Artefact,
+    architecture: str,
+) -> RowReturningQuery:
+    """
+    Helper method to get a query that fetches the latest Artefact Build IDs
```
omar-selo, line 69: latest or historic?

backend/test_observer/controllers/test_executions/logic.py
```diff
@@ -49,3 +58,124 @@ def parse_c3_test_result_status(status: C3TestResultStatus) -> TestResultStatus:
...
+        architecture (str): Architecture name to filter the results returned
+
+    Returns:
+        Prepared query object that can be executed as a query itself using .all()
+        or a subquery using .subquery()
+    """
+
+    return (
+        session.query(ArtefactBuild.id)
+        .distinct(ArtefactBuild.artefact_id)
```
omar-selo, line 87: does this produce a `DISTINCT` or `DISTINCT ON` statement?

backend/test_observer/controllers/test_executions/logic.py
andrejvelichkovski: Not sure what is a better naming option here. Maybe I should rename all methods/docstrings to previous, since they are previous to the current test execution.

omar-selo: previous also makes sense. But make sure to update the DTO object too

backend/test_observer/controllers/test_executions/logic.py
andrejvelichkovski: Depending on the DB dialect, for PostgreSQL it produces `DISTINCT ON`, for other SQL databases it produces standard `DISTINCT`.

Here is the query it produces:
```
SELECT DISTINCT ON (artefact_build.artefact_id) artefact_build.id 
FROM artefact_build JOIN artefact ON artefact.id = artefact_build.artefact_id 
WHERE artefact.name = 'core22' AND artefact.track = '22' AND artefact.repo IS NULL AND artefact.series IS NULL AND artefact_build.architecture = 'armhf' ORDER BY artefact_build.artefact_id DESC, artefact_build.revision DESC
```

omar-selo (approved):
Looks good to me!

PR #99, opened 2024-01-09
Use the web renderer

### LINKED ISSUE #71: no way to filter and ctrl+f does not work
I wanted to quickly find pc-kernel snap.

There is no filter to type `pc-kernel` to filter everything by pc-kernel.

I instead pressed `ctrl+f` and tried to search for pc-kernel, but somehow that doesn't work, as it brings up nothing.

mz2:
Use the web renderer so that for example finding text works.

Fixes https://github.com/canonical/test_observer/issues/71

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -16,7 +16,7 @@ COPY . /app
 
 RUN flutter pub get
 RUN flutter pub run build_runner build --delete-conflicting-outputs
-RUN flutter build web --release
+RUN flutter build web --release --web-renderer html
 
 FROM ubuntu:22.04
 ENV DEBIAN_FRONTEND=noninteractive


omar-selo (approved):
This is a very slight improvement. Namely, while ctrl+f works with this change, I found two limitations:
1. The search string can't have any special characters. (i.e. ' ', '_', ':', etc.) as it wouldn't work if it does
2. It can't find anything outside the view. So for instance if you need to scroll to find an artefact, then ctrl+f will not find it

These are very big limitations, specifically 2 would kind of render this useless. But I don't see an issue right now with making this change anyways.

nadzyah:
@omar-selo @mz2 shall we merge this PR?

omar-selo:
My concern is that html web renderer is [less performant](https://docs.flutter.dev/platform-integration/web/renderers#choosing-which-option-to-use). Given that it can't find text outside the field of view it's not adding much value. So if performance is an issue (as we can see now). Perhaps it's better not to merge this

omar-selo:
This is probably no longer needed given that we've added a [search box](https://github.com/canonical/test_observer/pull/122). So the underlying issue of finding is fixed

PR #104, opened 2024-01-12
Hot fix for user assignment

omar-selo:
Includes:
- Rename `user` table to `app_user` to avoid postgresql reserved name
- Commit db changes when assigning user to an artefact
- Add some users to seed script

### DIFF
--- a/backend/migrations/versions/2024_01_12_1040-bb2a51214402_.py
+++ b/backend/migrations/versions/2024_01_12_1040-bb2a51214402_.py
@@ -1,0 +1,28 @@
+"""empty message
+
+Revision ID: bb2a51214402
+Revises: c9851b127edc
+Create Date: 2024-01-12 10:40:00.596887+00:00
+
+"""
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "bb2a51214402"
+down_revision = "c9851b127edc"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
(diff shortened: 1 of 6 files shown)


backend/test_observer/controllers/users/users.py
```diff
@@ -11,7 +11,7 @@
 
 
 # include_in_schema=False to hide this endpoints from docs as its internal use
-@router.post("/", response_model=UserDTO, include_in_schema=False)
+@router.post("", response_model=UserDTO, include_in_schema=False)
```
nadzyah, line 14: sorry if it's a dumb question, but isn't `/` correct?

omar-selo: It's correct, but just to be consistent with other endpoints. The reasoning is described [here](https://github.com/canonical/test_observer/blob/main/backend/test_observer/main.py#L34-L39)

andrejvelichkovski (dismissed):
Looks good to me! Thanks for quickly addressing this issue!

backend/migrations/versions/2024_01_12_1040-bb2a51214402_.py
```diff
@@ -0,0 +1,28 @@
+"""empty message
```
nadzyah, line 1: would be nice to update the message here :)

nadzyah (dismissed):

andrejvelichkovski (approved):
Looks good to me

PR #106, opened 2024-01-15
Show assigned users on frontend

omar-selo:
Resolves [RTW-209](https://warthogs.atlassian.net/browse/RTW-209)

Changes:
- Show initials of users assigned to artefacts in artefact cards and dialog
- Show random [flutter material accent color](https://api.flutter.dev/flutter/material/Colors-class.html) depending on user id
- Show tooltip displaying user information when hovering over a user circle avatar

Screenshot from 2024-01-15 16-45-59
Screenshot from 2024-01-15 16-46-11

For reference the different color accents are all shown below (They're mostly different with a few slightly similar colors. But I felt it's fine):
Screenshot from 2024-01-15 16-45-34

[RTW-209]: https://warthogs.atlassian.net/browse/RTW-209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/lib/models/user.dart
+++ b/frontend/lib/models/user.dart
@@ -22,7 +22,7 @@ class User with _$User {
     final numOfNames = names.length;
     switch (numOfNames) {
       case 0:
-        return 'NA';
+        return 'N/A';
       case 1:
         return names[0][0].capitalize();
       default:
--- a/frontend/lib/models/user.dart
+++ b/frontend/lib/models/user.dart
@@ -1,0 +1,43 @@
+import 'package:dartx/dartx.dart';
+import 'package:flutter/material.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+part 'user.freezed.dart';
+part 'user.g.dart';
+
+Color generateColorFromString(String str) {
+  final hash = str.hashCode;
+  final r = ((hash >> 16) & 0xFF);
+  final g = ((hash >> 8) & 0xFF);
+  final b = (hash & 0xFF);
+  return Color.fromRGBO(r, g, b, 1.0);
+}
+
+@freezed
+class User with _$User {
+  const User._();
+
+  const factory User({
+    required int id,
+    required String name,
+    @JsonKey(name: 'launchpad_email') required String launchpadEmail,
+    @JsonKey(name: 'launchpad_handle') required String launchpadHandle,
+  }) = _User;
+
+  factory User.fromJson(Map<String, Object?> json) => _$UserFromJson(json);
+
+  String get initials {
+    final names = name.split(' ');
+    final numOfNames = names.length;
+    switch (numOfNames) {
+      case 0:
+        return 'NA';
+      case 1:
+        return names[0][0].capitalize();
+      default:
+        return names[0][0].capitalize() + names[numOfNames - 1][0].capitalize();
+    }
+  }
+
+  Color get color => generateColorFromString('$id');
+}
--- a/frontend/lib/models/user.dart
+++ b/frontend/lib/models/user.dart
@@ -1,17 +1,8 @@
 import 'package:dartx/dartx.dart';
-import 'package:flutter/material.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
 
 part 'user.freezed.dart';
 part 'user.g.dart';
-
-Color generateColorFromString(String str) {
-  final hash = str.hashCode;
-  final r = ((hash >> 16) & 0xFF);
-  final g = ((hash >> 8) & 0xFF);
-  final b = (hash & 0xFF);
-  return Color.fromRGBO(r, g, b, 1.0);
-}
 
 @freezed
 class User with _$User {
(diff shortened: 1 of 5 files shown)


andrejvelichkovski (approved):
Looks good to me! It's nice that Flutter provides different accent colours built in.

nadzyah:
IMO, these colours can definitely be seen as the same colour
image

frontend/lib/models/user.dart
```diff
@@ -0,0 +1,32 @@
...
+  }) = _User;
+
+  factory User.fromJson(Map<String, Object?> json) => _$UserFromJson(json);
+
+  String get initials {
+    final names = name.split(' ');
+    final numOfNames = names.length;
+    switch (numOfNames) {
+      case 0:
+        return 'NA';
```
nadzyah, line 25: does it mean "not assigned"? should we use "N/A" instead?

frontend/lib/ui/user_avatar.dart
```diff
@@ -0,0 +1,30 @@
...
+
+  final User user;
+
+  @override
+  Widget build(BuildContext context) {
+    return CircleAvatar(
+      backgroundColor: userIdToColor(user.id),
+      child: Tooltip(
+        message:
+            '${user.name}\n${user.launchpadEmail}\n${user.launchpadHandle}',
```
nadzyah, line 21: I suggest ordering it like `name, handle, email` like we have it on MM (skipping the titler ofc)
image

frontend/lib/models/user.dart
```diff
@@ -0,0 +1,32 @@
...
+    required String name,
+    @JsonKey(name: 'launchpad_email') required String launchpadEmail,
+    @JsonKey(name: 'launchpad_handle') required String launchpadHandle,
+  }) = _User;
+
+  factory User.fromJson(Map<String, Object?> json) => _$UserFromJson(json);
+
+  String get initials {
+    final names = name.split(' ');
+    final numOfNames = names.length;
```
nadzyah, line 22: could you please explain why we can have multiple names here?

nadzyah (commented):
Sorry if there are stupid questions. Your changes are great, there are just some things I'd like to understand better

frontend/lib/models/user.dart
omar-selo: As in first name, middle name, last name

nadzyah: cool, thanks!

andrejvelichkovski:
I agree with Nadzeya that these colours might be difficult to differentiate, but I'm also concerned that picking the primary colours might not be great choice, as the initials might be a bit difficult to read on a bright red colour.

Maybe we should try to find some colour palette that fits black well on top of it.

omar-selo:
> I agree with Nadzeya that these colours might be difficult to differentiate, but I'm also concerned that picking the primary colours might not be great choice, as the initials might be a bit difficult to read on a bright red colour.
> 
> Maybe we should try to find some colour palette that fits black well on top of it.
> 
> What do you think of [this palette](https://coolors.co/60695c-c9b6be-ffcf00-ee6123-35a7ff). The black text on it seems to be readable quite well.

That palette looks great. After discussing with Nadzeya we picked the following:
image

wdyt?

andrejvelichkovski:
This palette works well too, no need to change it.

frontend/lib/models/user.dart
```diff
@@ -0,0 +1,32 @@
...
+    switch (numOfNames) {
+      case 0:
+        return 'NA';
+      case 1:
+        return names[0][0].capitalize();
+      default:
+        return names[0][0].capitalize() + names[numOfNames - 1][0].capitalize();
```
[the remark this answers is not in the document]
nadzyah: hehe, a gen-Z inclusive approach :laughing:

nadzyah (approved):
+1, thanks for considering all my suggestions!

jocave:
> > I agree with Nadzeya that these colours might be difficult to differentiate, but I'm also concerned that picking the primary colours might not be great choice, as the initials might be a bit difficult to read on a bright red colour.
> > Maybe we should try to find some colour palette that fits black well on top of it.
> > What do you think of [this palette](https://coolors.co/60695c-c9b6be-ffcf00-ee6123-35a7ff). The black text on it seems to be readable quite well.
> 
> That palette looks great. After discussing with Nadzeya we picked the following: image
> 
> wdyt?

I don't believe it is incredibly important to distinguish the colours as they do not convey meaning - e.g. if the colour indicated whether the artefact was signed off then the should be differentiable by as many people as possible. This is just categorisation and the meaning can be discerned by the letters. The more important part is indeed being able to read the chars on the background colour.

nadzyah:
> I don't believe it is incredibly important to distinguish the colours as they do not convey meaning - e.g. if the colour indicated whether the artefact was signed off then the should be differentiable by as many people as possible. This is just categorisation and the meaning can be discerned by the letters. The more important part is indeed being able to read the chars on the background colour.

I believe all of these colours are fine to be used as background for black chars

frontend/lib/models/user.dart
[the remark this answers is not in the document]
omar-selo: Looks like it works. Basically it just returns that character (emoji)

PR #112, opened 2024-01-23
Admin actions including delete artefact

omar-selo:
Resolves [RTW-216](https://warthogs.atlassian.net/browse/RTW-216). Basically we needed an admin action to delete an artefact. After looking around it seems that the best approach is to use a charm action as we'd be able to pass in database url environment variable that way. But this charm action can't directly import a function from our backend code as it's a different package. Instead, it can run a script that does the same operation.

Changes:
- Add delete artefact script and action
- Move add user endpoint to become a script and an action
- Move change assignee endpoint to become a script and an action

One open question is where should we add documentation about admin actions? Some options:
1. Add a new file about admin or operations to [TO documentation directory on google drive](https://drive.google.com/drive/folders/16ZxIUikANEYl4MVdXqkoTiq5cbqOAAGJ) 
2. Add a heading about admin tasks in main readme of this repo
3. Add a heading about admin tasks in backend readme of this repo
4. Add a readme under `test_observer/backend/charm` with this info


backend/scripts/add_user.py
```diff
@@ -0,0 +1,16 @@
+from argparse import ArgumentParser
```
nadzyah, line 1: Since this is a script, I guess it’s better to add `#!/usr/bin/python3` or `#!/usr/bin/env python3`

nadzyah: Same in all the following scripts

andrejvelichkovski:
This looks good overall, I was just curious how do we trigger these actions, is it through the CLI or there is some other way.

In terms of the documentation, I think we can rename the "Debug TO" to "TO maintenance" and we can store all documentation about maintaining TO over there, such as running these admin actions, checking the database and similar operations, to avoid create more files and places where we store the documentation.

omar-selo:
> how do we trigger these actions, is it through the CLI or there is some other way

Yes we'd run them through CLI. An example would be `juju run-action api/leader change-assignee artefact-id=1 user-id=1 --wait`.



> I think we can rename the "Debug TO" to "TO maintenance" and we can store all documentation about maintaining TO over there

I don't mind this. Sounds reasonable to me.

andrejvelichkovski:
Since TO is a new project and we are just starting to add new tasks, I wanted to suggest we make the CLI tools using the python [click](https://click.palletsprojects.com/en) library instead of argparse. It comes with a few nice helpers out-of-the-box and simplifies the writing process.

Our tools are quite simple for now, but might be nice to have a good tool in case we decide to extend them in the future and add some more advanced functionalities.

backend/scripts/seed_data.py
```diff
@@ -1,30 +1,27 @@
...
     C3TestResultStatus,
     EndTestExecutionRequest,
     StartTestExecutionRequest,
 )
-from test_observer.controllers.users.models import CreateUserRequest
 from test_observer.data_access.models_enums import FamilyName
+from test_observer.users.add_user import add_user
 
 BASE_URL = "http://localhost:30000/v1"
 CREATE_USER_URL = f"{BASE_URL}/users"
```
andrejvelichkovski, line 21: Just a small nitpick, maybe we can remove this variable as it is no longer used.

andrejvelichkovski (approved):
Looks good to me overall, I just had one nitpick, but feel free to ignore it.

I can re-review if you decide to switch to the click library, or we can go ahead and deploy the feature.

backend/scripts/seed_data.py
omar-selo: I'm surprised that ruff didn't complain about it

omar-selo:
> I wanted to suggest we make the CLI tools using the python [click](https://click.palletsprojects.com/en) library instead of argparse.

Click looks like a nice library frankly. But given that the scripts will be mainly used by charm subprocess instead of humans, it seems like an overkill to me (we'd only use the scripts directly in development).

andrejvelichkovski (approved):
Looks good to me!

andrejvelichkovski:
> > I wanted to suggest we make the CLI tools using the python [click](https://click.palletsprojects.com/en) library instead of argparse.
> 
> Click looks like a nice library frankly. But given that the scripts will be mainly used by charm subprocess instead of humans, it seems like an overkill to me (we'd only use the scripts directly in development).

Fair enough, it might be an overkill for now.

PR #117, opened 2024-01-29
Add frontend filters

omar-selo:
Resolves [RTW-218](https://warthogs.atlassian.net/browse/RTW-218).

Changes:
- Add assignee and status filters
- Require dart 3 and above to allow the use of dart records (used in filter options)
- Adds the first real test cases and includes some refactor and setup to get them working
- Add a basic and overdue README

Notes:
- For UI design I was inspired by [ubuntu certified page](https://ubuntu.com/certified?q=&limit=20)
- Tests now only work on chrome because we import `dart:html`
- I've only added assignee and status filters because those are the only ones that made sense to me. But it's is trivial to add more if some other filters are useful

Video:
[Screencast from 2024-01-29 16-35-32.webm](https://github.com/canonical/test_observer/assets/4087200/179dab06-ed1f-4e37-9542-c6cb909893a6)

[RTW-218]: https://warthogs.atlassian.net/browse/RTW-218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -41,4 +41,4 @@ $ flutter test --platform chrome # requires chrome
 
 ## Connect to backend
 
-By default TO frontend will use `http://localhost:30000/` to communicate with backend. To use a different address you can modify `window.testObserverAPIBaseURI` attribute in `web/index.html`.
+By default the Test Observer frontend will use `http://localhost:30000/` to communicate with the backend. To use a different address you can modify `window.testObserverAPIBaseURI` attribute in `web/index.html`.
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -1,2 +1,44 @@
-# testcase_dashboard
-Flutter based dashboard for visualising SRU regression test results
+# Test Observer Frontend
+
+The frontend of TO developed in Flutter but meant to be a web application.
+
+## Setup
+
+Follow official flutter installation [instructions](https://docs.flutter.dev/get-started/install/linux).
+
+Then install dependencies via:
+```bash
+$ flutter pub get
+```
+
+## Run
+
+This project uses code generation, so you must first generate the required code:
+
+```bash
+$ dart run build_runner build
+```
+
+Then you can launch the fontend on chrome:
+
+```bash
+$ flutter run -d chrome
+```
+
+## Test
+
+### Static analysis
+
+```bash
+$ flutter analyze
+```
+
+### Run tests
+
+```bash
+$ flutter test --platform chrome # requires chrome
+```
+
+## Connect to backend
+
+By default TO frontend will use `http://localhost:30000/` to communicate with backend. To use a different address you can modify `window.testObserverAPIBaseURI` attribute in `web/index.html`.
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -1,6 +1,6 @@
 # Test Observer Frontend
 
-The frontend of TO developed in Flutter but meant to be a web application.
+The frontend of Test Observer is a web application developed in Flutter.
 
 ## Setup
 
--- a/frontend/lib/models/filter.dart
+++ b/frontend/lib/models/filter.dart
@@ -1,0 +1,8 @@
+import 'artefact.dart';
+
+class Filter {
+  final String name;
+  final bool Function(Artefact) shouldInclude;
+
+  const Filter({required this.name, required this.shouldInclude});
+}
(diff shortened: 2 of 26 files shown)


frontend/README.md
```diff
@@ -1,2 +1,44 @@
...
+
+### Run tests
+
+```bash
+$ flutter test --platform chrome # requires chrome
+```
+
+## Connect to backend
+
+By default TO frontend will use `http://localhost:30000/` to communicate with backend. To use a different address you can modify `window.testObserverAPIBaseURI` attribute in `web/index.html`.
```
plars, line 44:
```suggestion
By default the Test Observer frontend will use `http://localhost:30000/` to communicate with the backend. To use a different address you can modify `window.testObserverAPIBaseURI` attribute in `web/index.html`.
```

frontend/README.md
```diff
@@ -1,2 +1,44 @@
-# testcase_dashboard
-Flutter based dashboard for visualising SRU regression test results
+# Test Observer Frontend
+
+The frontend of TO developed in Flutter but meant to be a web application.
```
plars, line 3:
```suggestion
The frontend of Test Observer is a web application developed in Flutter.
```

plars (commented):
admittedly, the dart code here is a bit opaque to me but I really appreciate the setup/test instructions in the README, that's a great addition, and I intend to give it a try!

Other than a few suggestions for improved wording in the README that I found, I was wondering how big of a change it would be to have the filter widget could show/hide with a button so that it doesn't always take up space if you don't want it?

frontend/lib/models/filter.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+import 'artefact.dart';
+
+part 'filter.freezed.dart';
+
+@freezed
+class Filter with _$Filter {
```
andrejvelichkovski, line 8: Just a small suggestion, maybe ArtefactFilter would be a better name here, as this filter only works for Artefacts.

andrejvelichkovski (dismissed):
Overall, this PR looks good to me and I'm happy for it to be merged. My only idea would be to possibly make the filter horizontal rather than vertical, under the "Snap/Deb Update Verification". I think it would make it a bit less packed on smaller screens that can't fit 5 columns of data.

omar-selo:
The issue of filters taking space is very valid. Indeed, after investigating how this would look at with different resolutions, I found that I can make small changes to fit a 1920x1080 resolution but not smaller. So here are the options so far:
a. Assume that 1920x1080 is the smallest resolution we care about (easiest)
b. Show and hide the filters. We can play around with where the button is and whether the filters come from the right or left and whether they appear on top of the board or just push the board. But in general with this approach we need to figure out a way to make the button obvious (not hidden):
[Screencast from 2024-01-30 12-51-56.webm](https://github.com/canonical/test_observer/assets/4087200/199e2dca-fe16-4640-88dc-5743dd84ed11)
c. Show filters above the dashboard horizontally. We will probably need to change how filters are displayed though with this approach. One idea is to bake it in a search bar. Vanilla design has a nice approach to this [here](https://vanillaframework.io/docs/examples/patterns/search-and-filter/expanded). Once some filters are selected they would be shown as chips like [this](https://vanillaframework.io/docs/examples/patterns/search-and-filter/with-chips). The issue with this however is that I didn't find a flutter widget that implements this special search box. So we'd need to implement it ourselves which seems like it would take some time. A simpler approach to putting filters horizontally is to have each filter be a drop down menu that opens up the options and you can click on the checkbox. But that doesn't sound like a standard design used elsewhere.

omar-selo:
I guess then we can add the button now like this:
[Screencast from 2024-01-30 13-46-25.webm](https://github.com/canonical/test_observer/assets/4087200/7bf969cb-9863-4e45-a7e8-0d468db6a278)

Later on when we have a search bar we can move it next to the search bar. If users can't see this button then we'll try to make it more visible.

andrejvelichkovski (approved):
Thanks for quickly exploring different options on how to make it fit smaller screens. This PR now looks good to me and I think it's ready to be landed.

plars (approved):
Indeed, something like that button to show/hide the filters is exactly what I was hoping for! that helps a lot with the horizontal space usage. Looks good, +1

PR #119, opened 2024-01-30
Make end test idempotent

omar-selo:
Resolves [RTW-222](https://warthogs.atlassian.net/browse/RTW-222). Additionally, I resolved warnings we keep getting when running pytest. Sadly, I didn't find a way to resolve those warnings without modifying the code itself.

[RTW-222]: https://warthogs.atlassian.net/browse/RTW-222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/test_observer/controllers/test_executions/logic.py
+++ b/backend/test_observer/controllers/test_executions/logic.py
@@ -26,10 +26,9 @@ def reset_test_execution(
     test_execution.c3_link = None
     test_execution.review_decision = []
     test_execution.review_comment = ""
-    db.execute(
-        delete(TestResult).where(TestResult.test_execution_id == test_execution.id)
-    )
     db.commit()
+
+    delete_previous_results(db, test_execution)
 
 
 def compute_test_execution_status(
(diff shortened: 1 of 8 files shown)


backend/test_observer/controllers/test_executions/logic.py
```diff
@@ -13,7 +13,23 @@
...
+    test_execution: TestExecution,
+):
+    test_execution.status = TestExecutionStatus.IN_PROGRESS
+    test_execution.ci_link = request.ci_link
+    test_execution.c3_link = None
+    test_execution.review_decision = []
+    test_execution.review_comment = ""
+    db.execute(
+        delete(TestResult).where(TestResult.test_execution_id == test_execution.id)
+    )
```
andrejvelichkovski, line 31: Should we use the `delete_previous_results` function here, to make sure we only maintain one place where we delete TestResults.

andrejvelichkovski (changes requested):
Thanks for these fixes. The warnings in the tests were really annoying. The code looks good to me overall, but I just have one improvement idea.

andrejvelichkovski (approved):
Looks good to me

PR #123, opened 2024-02-05
Add frontend integration tests to ci

omar-selo:
Managed to get frontend integration tests working on GitHub workflow.

### DIFF
--- a/.github/workflows/test_frontend.yml
+++ b/.github/workflows/test_frontend.yml
@@ -25,10 +25,18 @@ jobs:
       - run: flutter pub get
       - run: flutter pub run build_runner build
       - run: flutter analyze
-      - name: Install Google Chrome
-        run: |
-          wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
-          sudo apt update -y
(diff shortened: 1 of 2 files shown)


andrejvelichkovski (approved):
LGTM! Just curious if we should try to commit one failing test (maybe assert 1 == 2) to see that the job indeed fails if some integration test fails.

omar-selo:
> LGTM! Just curious if we should try to commit one failing test (maybe assert 1 == 2) to see that the job indeed fails if some integration test fails.

Ah don't worry. Actually, turns out the integration test was wrong because I wasn't using the mocked `ApiReporistory`. I found that out via this workflow, so it's definitely working.

andrejvelichkovski:
> > LGTM! Just curious if we should try to commit one failing test (maybe assert 1 == 2) to see that the job indeed fails if some integration test fails.
> 
> Ah don't worry. Actually, turns out the integration test was wrong because I wasn't using the mocked `ApiReporistory`. I found that out via this workflow, so it's definitely working.

That's a proof it works for sure :)

PR #125, opened 2024-02-16
Artefact status changes guards

omar-selo:
Resolves [RTW-215](https://warthogs.atlassian.net/browse/RTW-215).

Changes:
- Adds backend validation checks that reject:
  - approving artefacts that have unapproved test executions
  - rejecting artefacts that have no rejected test executions
- Add global api error handling on frontend. Such that if any api request gets a meaningful error response message, a popup would show that error. See video below. I decided to go with this approach as it's easy to support new errors. All we have to do is raise an HTTPException from the backend and it would automatically show as a popup on the frontend. I definitely do not want to repeat the same logic/checks on the frontend. And so far I don't think it's necessary to have a custom way to respond to individual errors. (although I could be wrong)

Notes:
- These status change guards are not constraints, just checks. It is possible to approve an artefact then change the review status of a test execution to be rejected or undecided. This is desired behavior (from my discussion with @jocave). That's because signoffs should be ideally a final say, and shouldn't change easily without manual intervention.

[Screencast from 2024-02-16 14-33-31.webm](https://github.com/canonical/test_observer/assets/4087200/1876c491-2c0a-4066-8d9e-2e3c45073505)

[RTW-215]: https://warthogs.atlassian.net/browse/RTW-215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/test_observer/controllers/artefacts/artefacts.py
```diff
@@ -70,14 +74,36 @@ def get_artefact(artefact_id: int, db: Session = Depends(get_db)):
...
+        )
+
+    if (
+        request.status == ArtefactStatus.MARKED_AS_FAILED
+        and not is_there_a_rejected_test_execution(artefact)
+    ):
+        raise HTTPException(
+            400,
+            detail="At least one test execution needs to be rejected",
+        )
```
andrejvelichkovski, line 103: Just a small code style suggestion, maybe we can extract this into helper `_validate_request_status` to clear up the main login of this method.

andrejvelichkovski (dismissed):
Overall looks good to me. Just one style comment, but feel free to skip it if you don't like that style :D

The use of pop-up for errors should be good for now, as most of the front-end is read-only, with a small input interaction from the user.

backend/test_observer/controllers/artefacts/artefacts.py
```diff
@@ -84,28 +84,31 @@ def patch_artefact(
     if artefact is None:
         raise HTTPException(status_code=404, detail="Artefact not found")
 
-    if (
-        request.status == ArtefactStatus.APPROVED
-        and not are_all_test_executions_approved(artefact)
+    _validate_artefact_status(request, artefact)
```
andrejvelichkovski, line 87: Shouldn't we pass the parameters in the opposite order, and to pass request.status instead of request: `_validate_artefact_status(artefact, request.status)`

andrejvelichkovski (changes requested):

andrejvelichkovski (approved):

backend/test_observer/controllers/artefacts/artefacts.py
andrejvelichkovski: It happens :)

PR #126, opened 2024-02-19
Include c3 link in the request body of the end-test API endpoint

andrejvelichkovski:
Currently to send the C3 link we make an additional PATCH request to the `/test-execution/ID` endpoint. This introduces some extra network overhead, which is not necessary, as we already have the communication from C3 to TO with the `end-test/ID` endpoint. It resolves [RTW-240](https://warthogs.atlassian.net/browse/RTW-240).

This implementation adds the C3 link as an optional parameter in the request body of the `end-test/ID/` endpoint. It is optional to ensure backwards compatibility with all clients making use of this API without sending the c3 link in the request body.

[RTW-240]: https://warthogs.atlassian.net/browse/RTW-240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/test_observer/controllers/test_executions/models.py
+++ b/backend/test_observer/controllers/test_executions/models.py
@@ -99,6 +99,7 @@ class C3TestResult(BaseModel):
 class EndTestExecutionRequest(BaseModel):
     id: int
     ci_link: Annotated[str, HttpUrl]
+    c3_link: Annotated[str, HttpUrl] | None = None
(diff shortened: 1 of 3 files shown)


omar-selo (approved):
LGTM!

PR #127, opened 2024-02-20
Add testresult report

omar-selo:
Resolves [RTW-228](https://warthogs.atlassian.net/browse/RTW-228).

Changes:
- Adds a new endpoint `/v1/reports/testresults` that returns a CSV file with information about test results (we can then use this csv file to find out interesting information like the tests that fail most)
- Make track, store, series and repo fields in artefact not nullable as it doesn't make sense to have two values that mean the same thing (problem was apparent in the report)
- Created a new `DataGenerator` class to help with generating dummy data for tests. Refactored existing `create_artefact` and `create_user` functions into it

Note:
- I didn't refactor all tests to use `DataGenerator` to create all of the objects yet, as that would make this PR too big. But can definitely do that in a future PR

Example CSV report file:
[testresults.csv](https://github.com/canonical/test_observer/files/14345986/testresults.csv)

[RTW-228]: https://warthogs.atlassian.net/browse/RTW-228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plars (commented):
The main question I have, is "why csv and not json?"

andrejvelichkovski (changes requested):
The PR looks good to me, but I'm just concern about the NULLs in the migration. Thanks for refactoring the logic for creating all entities in a separate data generator, it really simplifies the process.

For the refactoring of existing tests, maybe we can start requesting it from each other every time we make a change to an existing test or add new one?

omar-selo:
> The main question I have, is "why csv and not json?"

Adding to @andrejvelichkovski's point, it's also much easier to play around with this data via some spreadsheet software. We can filter, sort, create charts, etc. With json we'd have to do these operations through code (unless there is a tool out there that makes this easier to work with through GUI).

andrejvelichkovski (approved):
Thanks for quickly following up. LGTM!

andrejvelichkovski (approved):
LGTM, I think just one follow-up we can later do for the `null` clean-up would be to remove any Nulls sent from the Jenkins jobs.

PR #129, opened 2024-02-21
Ignore old builds when validating if artefact status change is allowed

omar-selo:
Hot fix for the issue discussed [here](https://chat.canonical.com/canonical/pl/ttrjdfmcdpbodeyomwxkpgsbcy). Basically, when checking if an artefact status change is allowed or not we shouldn't take into consideration old hidden builds.


nadzyah (approved):

andrejvelichkovski (approved):
The changes make sense to me. I'm just curious whether taking the maximum of revision will work for debs, as I thought we don't use revisions for debs.

PR #130, opened 2024-02-23
Only consider latest builds when checking if signoff is allowed

omar-selo:
Looks like the previous [PR](https://github.com/canonical/test_observer/pull/129) to fix the issue of "when checking if an artefact status change is allowed or not we shouldn't take into consideration old hidden builds." doesn't work. Main reason is that turns out `itertools.groupby` requires input to be sorted. This PR should fix the issue, and I've consolidated this logic in one place.

### DIFF
--- a/backend/test_observer/data_access/queries.py
+++ b/backend/test_observer/data_access/queries.py
@@ -1,0 +1,17 @@
+from sqlalchemy import Select, select
+
+from test_observer.data_access.models import ArtefactBuild
+
+latest_artefact_builds = (
+    select(ArtefactBuild)
+    .distinct(ArtefactBuild.artefact_id, ArtefactBuild.architecture)
+    .order_by(ArtefactBuild.artefact_id, ArtefactBuild.architecture)
+)
+
+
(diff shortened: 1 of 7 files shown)


backend/test_observer/data_access/queries.py
andrejvelichkovski, line 1: It seems like now we are dividing the database helper functionalities between `repository` and `data_access`. Do you think we can maybe move the new `data_access` functionalities into `repository`, or the difference in them is bigger?

I really like the idea that we will use the same query/function to fetch and verify the latest submissions per artefact, it should be less error-prone.

omar-selo: So right now we can still proceed with two files. Repository is for functions that return final results. Queries is for functions/variables that return queries that need to be run. If we find that queries approach is really better, we can move repository functions into queries then delete that file. wdyt?

andrejvelichkovski: That makes sense, we can land this as it is, and we can explore different options on how to structure the project in some common patterns or architectures and maybe as we add new functionalities we can shape them towards that way.

PR #132, opened 2024-02-29
Kernel sru integration

omar-selo:
Resolves the backend part of [RTW-211](https://warthogs.atlassian.net/browse/RTW-211).

Changes:
- Add bug link field to artefact
- Add logic to integrate with kernel swm. This is now different from how it worked [before](https://github.com/canonical/certification-dashboard-manager/blob/main/dashboard_manager/core/kernel_sru_sync.py). Namely, the swm json now references our artefact id and also includes due date. So we just get it from them. These additions were thanks to Andy Whitcroft from the kernel team.
- Add a celery task that periodically does this integration every 5 minutes.
- Return from api bug link and due date.

Testing:
- Apart from unit tests, I've also tested that the job does run periodically and update bug link and due date.

[RTW-211]: https://warthogs.atlassian.net/browse/RTW-211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -27,7 +27,7 @@ Linting is done using ruff, formatting using black and type checking using mypy.
 
 ### 4. Start the development environment
 
-Assuming that your microk8s cluster is running, you can start the development environment by simply running `$ skaffold dev`. This command will build the docker images and push them to your microk8s registry, then apply your k8s manifest to start the cluster and pull those images. Additionally, skaffold will watch for file changes and either sync them directly inside the running containers or rebuild and redeploy k8s cluster for you automatically.
+Assuming that your microk8s cluster is running, you can start the development environment by simply running `$ skaffold dev --no-prune=false --cache-artifacts=false`. This command will build the docker images and push them to your microk8s registry, then apply your k8s manifest to start the cluster and pull those images. Additionally, skaffold will watch for file changes and either sync them directly inside the running containers or rebuild and redeploy k8s cluster for you automatically.
 
 ### 5. [Optional] seed the database
 
@@ -75,7 +75,7 @@ Since the database is in microk8s cluster, migrations have to be applied inside 
 
 ## Tests
 
-To run the tests, first make sure that you're running `skaffold dev` or `skaffold run`. Then execute the pytest command in the API pod:
+To run the tests, first make sure that you're running skaffold. Then execute the pytest command in the API pod:
 
 `$ kubectl exec -it service/test-observer-api -- pytest`
 
--- a/backend/test_observer/controllers/artefacts/models.py
+++ b/backend/test_observer/controllers/artefacts/models.py
@@ -17,6 +17,8 @@
 # Written by:
 #        Omar Selo <omar.selo@canonical.com>
 #        Nadzeya Hutsko <nadzeya.hutsko@canonical.com>
+from datetime import date
+
 from pydantic import AliasPath, BaseModel, ConfigDict, Field
 
 from test_observer.data_access.models_enums import (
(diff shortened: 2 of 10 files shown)


backend/README.md
```diff
@@ -27,7 +27,7 @@ Linting is done using ruff, formatting using black and type checking using mypy.
 
 ### 4. Start the development environment
 
-Assuming that your microk8s cluster is running, you can start the development environment by simply running `$ skaffold dev`. This command will build the docker images and push them to your microk8s registry, then apply your k8s manifest to start the cluster and pull those images. Additionally, skaffold will watch for file changes and either sync them directly inside the running containers or rebuild and redeploy k8s cluster for you automatically.
+Assuming that your microk8s cluster is running, you can start the development environment by simply running `$ skaffold dev --no-prune=false --cache-artifacts=false`. This command will build the docker images and push them to your microk8s registry, then apply your k8s manifest to start the cluster and pull those images. Additionally, skaffold will watch for file changes and either sync them directly inside the running containers or rebuild and redeploy k8s cluster for you automatically.
```
andrejvelichkovski, line 30: It's a shame `skaffold` spells it as `artifacts` :smile:

backend/test_observer/controllers/artefacts/models.py
```diff
@@ -46,6 +48,8 @@ class ArtefactDTO(BaseModel):
     stage: str = Field(validation_alias=AliasPath("stage", "name"))
     status: ArtefactStatus
     assignee: UserDTO | None
+    due_date: date | None
+    bug_link: str
```
andrejvelichkovski, line 52: For other links we use `Annotated[str, HttpUrl]`, should we maybe do same here?

backend/test_observer/kernel_swm_integration/swm_reader.py
```diff
@@ -0,0 +1,56 @@
...
+def _is_tracker_open(tracker: dict) -> bool:
+    closed_statuses = ("Fix Committed", "Fix Released")
+    status = tracker.get("task", {}).get("kernel-sru-workflow", {}).get("status")
+    return status and status not in closed_statuses
+
+
+def _extract_artefact_id(tracker: dict) -> int | None:
+    to_info = tracker.get("test-observer")
+    if to_info:
+        possible_keys = ("edge", "beta", "candidate", "stable", "proposed", "updates")
```
andrejvelichkovski, line 45: Should we maybe the `possible_keys` variable in a separate constants file, to make it easier to change in case we start testing more things?

andrejvelichkovski (dismissed):
Looks good to me! Only left two style comments

backend/README.md
omar-selo: actually that makes it easier to make the distinction between our artefacts and everyone else's artifacts :smiley:

backend/test_observer/controllers/artefacts/models.py
omar-selo: Those are requests, while this is a response. While we want to validate data going into our application to make sure it's clean, we probably shouldn't validate data going out of the application. Otherwise if say an artefact for some reason had a malformed link, then it would just never be returned as that endpoint would fail

andrejvelichkovski: Ah good point, please ignore it.

backend/test_observer/kernel_swm_integration/swm_reader.py
omar-selo: Good idea, I'll actually just read them from the database

PR #135, opened 2024-03-04
Make artefact dialog a page instead

omar-selo:
Resolves [RTW-242](https://warthogs.atlassian.net/browse/RTW-242).

image

[RTW-242]: https://warthogs.atlassian.net/browse/RTW-242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontend/lib/ui/artefact_page/artefact_dialog_info_section.dart
andrejvelichkovski, line 1: Should we rename this page as well if we are getting rid of `artefact_dialog` completely

omar-selo: Ah yes good catch, I missed that

frontend/lib/ui/artefact_page/artefact_page_header.dart
andrejvelichkovski, line 1: This file seems to be the same content as the old `artefact_dialog_header`, just curious why GitHub didn't pick it up as a rename rather than deletion and a new file.

frontend/lib/ui/artefact_page/artefact_page.dart
```diff
@@ -0,0 +1,65 @@
...
+
+class ArtefactPage extends ConsumerWidget {
+  const ArtefactPage({super.key, required this.artefactId});
+
+  final int artefactId;
+
+  Column get _invalidArtefactErrorMessage {
+    return const Column(
+      children: [
+        DialogHeader(),
```
andrejvelichkovski, line 21: Given that we don't have a Dialog anymore, do we still need to render the error message in a Dialog box, or we can just maybe render it on the page itself?

andrejvelichkovski (approved):
Approving as overall it looks good to me, but I'm not sure if the dialog box is still what we want to show when the artefact is invalid?

frontend/lib/ui/artefact_page/artefact_page.dart
omar-selo: UX wise, I imagine there are two kinds of errors. First is when a user was supposed to navigate to a certain page but something went wrong. Second is when a user is just making some kind of action on the same page without navigating. I think it makes sense in the first case to navigate to an error page. But in the second case perhaps it's more user friendly to just have a dialog. wdyt?

PR #137, opened 2024-03-06
Show text message when test results are not available

andrejvelichkovski:
This PR resolves RTW-247. It adds a condition on whether the test execution is completed (failed or passed) in which case it shows the test filters (Passed / Skipped / Failed) otherwise it shows a text that 'This test execution is not completed. No test results are available yet.'.

Only modifications on the frontend are performed.

### Screenshot of the modified artefact page

Screenshot from 2024-03-06 09-33-43

### DIFF
--- a/frontend/lib/models/test_execution.dart
+++ b/frontend/lib/models/test_execution.dart
@@ -51,6 +51,18 @@ enum TestExecutionStatus {
   @JsonValue('PASSED')
   passed;
 
+  bool get isCompleted {
+    switch (this) {
+      case notStarted:
+      case inProgress:
+      case notTested:
+        return false;
+      case passed:
+      case failed:
+        return true;
+    }
+  }
+
   String get name {
     switch (this) {
       case notStarted:
--- a/frontend/lib/ui/artefact_page/test_execution_expandable.dart
+++ b/frontend/lib/ui/artefact_page/test_execution_expandable.dart
@@ -50,14 +50,27 @@ class TestExecutionExpandable extends ConsumerWidget {
           ),
         ],
       ),
-      children: TestResultStatus.values
-          .map(
-            (status) => TestResultsFilterExpandable(
-              statusToFilterBy: status,
-              testExecutionId: testExecution.id,
-            ),
-          )
-          .toList(),
+      children: (!testExecution.status.isCompleted)
+          ? [
+              Align(
+                alignment: Alignment.topLeft,
+                child: Padding(
+                  padding: const EdgeInsets.all(8.0),
+                  child: Text(
+                    'This test execution is not completed. No test results are available yet.',
+                    style: Theme.of(context).textTheme.bodyLarge,
+                  ),
+                ),
+              ),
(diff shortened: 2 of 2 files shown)


frontend/lib/models/test_execution.dart
```diff
@@ -51,6 +51,18 @@ enum TestExecutionStatus {
...
+    switch (this) {
+      case notStarted:
+      case inProgress:
+      case notTested:
+        return false;
+      case passed:
+      case failed:
+        return true;
+    }
+  }
```
omar-selo, line 64: nice

omar-selo (commented):
I'm wondering if this solution is any better. As in the user still sees that there is a drop down and only after opening the drop down they know that there's nothing under it.

andrejvelichkovski:
I have changed the design to fit more with the style of what we already have for completed Test Executions. Namely, now it is a sort-of disable expansion tile (standard ListTile), which has the same consistency with the standard ExpansionTiles (for completed environments).

Screenshot from 2024-03-06 12-06-51
Screenshot from 2024-03-06 12-07-05

frontend/lib/ui/artefact_page/test_execution_expandable.dart
```diff
@@ -13,43 +13,56 @@ class TestExecutionExpandable extends ConsumerWidget {
 
   final TestExecution testExecution;
 
-  @override
-  Widget build(BuildContext context, WidgetRef ref) {
+  Row getTestExecutionTileTitle(BuildContext context) {
```
omar-selo, line 16: small thing, but perhaps this could be a widget on it's own

frontend/lib/ui/artefact_page/test_execution_expandable.dart
```diff
@@ -13,43 +13,56 @@ class TestExecutionExpandable extends ConsumerWidget {
...
+            const SizedBox(width: Spacing.level3),
+            if (c3Link != null)
+              InlineUrlText(
+                url: c3Link,
+                urlText: 'C3',
+              ),
+          ],
+        ),
+      ],
+    );
```
omar-selo, line 48: it just occurred to me that there is a row in a row for no apparent reason

omar-selo (approved):
Overall the changes look good. I added a couple of comments, but feel free to ignore them if you disagree

frontend/lib/ui/artefact_page/test_execution_expandable.dart
andrejvelichkovski: Good idea, I converted into a small private widget.

frontend/lib/ui/artefact_page/test_execution_expandable.dart
andrejvelichkovski: I guess we have added it at some point to logically separate the left and right components, but I agree there is no apparent reason to have it, I have deleted it.

omar-selo (approved):
LGTM!

PR #138, opened 2024-03-06
Adds filters for test executions

omar-selo:
First PR towards [RTW-244](https://warthogs.atlassian.net/browse/RTW-244).

Changes:
- Adds side filters in artefact page that enable the user to filter test executions by status and review status.
- Refactor artefacts filters logic to be able to common code/logic for test executions too (DRY).

Video:
[Screencast from 2024-03-06 16-42-42.webm](https://github.com/canonical/test_observer/assets/4087200/48c3b534-cf9f-4412-8897-b378ef998325)

[RTW-244]: https://warthogs.atlassian.net/browse/RTW-244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andrejvelichkovski:
Just curious, do you think it will be better if we also put the button to hide/show the filters here as well, similar to how we do it on the main page for filtering artefacts?

frontend/lib/models/filter.dart
```diff
@@ -0,0 +1,55 @@
...
+    required List<({String name, bool value})> options,
+  }) = _Filter<T>;
+
+  factory Filter.fromObjects({
+    required String name,
+    required String? Function(T) extractOption,
+    required List<T> objects,
+  }) {
+    final names = <String>{};
+    for (final artefact in objects) {
```
andrejvelichkovski, line 22: Sorry to be really nitpicking here, but this is a bit confusing as it is not really iterating over artefacts, but any type of object that uses this filter, if you don't mind we can change it to `for (final object in objects)`

frontend/lib/ui/dashboard/artefact_side_filters.dart
```diff
@@ -0,0 +1,37 @@
...
+
+    return SizedBox(
+      width: SideFilters.width,
+      child: Column(
+        children: [
+          ArtefactSearchBar(),
+          const SizedBox(height: SideFilters.spacingBetweenFilters),
+          SideFilters(
+            filters: filters,
+            onOptionChanged: (filterName, optionName, value) => ref
```
andrejvelichkovski, line 25: Small detail here, but should we pass the .handleFilterOptionChange function without any parameters like we do for the TestExecutionSideFilter to keep it more consistent and simpler?

frontend/lib/providers/filtered_test_execution_ids.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:dartx/dartx.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+import 'artefact_builds.dart';
+import 'test_execution_filters.dart';
+
+part 'filtered_test_execution_ids.g.dart';
+
+@riverpod
+Set<int> filteredTestExecutionIds(
```
andrejvelichkovski, line 10: Just curious, does it make any difference that here we use `Set<int>` instead of `Set<TestExecution>` that way we wouldn't need to convert the IDs back to test executions in the widget code?

andrejvelichkovski (dismissed):
Really nice that you managed to fully avoid duplicating any code and it's nice to see this working! Overall looks good to me just had a few small details, feel free to ignore them if you don't see them fit.

andrejvelichkovski:
> > Just curious, do you think it will be better if we also put the button to hide/show the filters here as well, similar to how we do it on the main page for filtering artefacts?
> 
> I opted to not put a button here since there is so much space. Unlike the dashboard that has 4 columns already. This page has no problem fitting things horizontally even on smaller screens.

That makes sense! Thanks for clarifying it

frontend/lib/models/filter.dart
omar-selo: Ah good catch, I missed this

frontend/lib/ui/dashboard/artefact_side_filters.dart
omar-selo: Another good catch

frontend/lib/providers/filtered_test_execution_ids.dart
omar-selo: Issue here is that to keep the same current UI we need to iterate over test executions per artefact build. So having a list or set of test executions wont help. Instead, we have to iterate over artefact builds and check each test execution inside it. I considered moving artefact builds into test executions and changing the UI to just show test executions. But I figured that's too much change to include in this PR

andrejvelichkovski: Ah I see, I forgot about the existance of ArtefactBuilds in the UI, we can keep it as it is now.

andrejvelichkovski (approved):
Thanks for quickly iterating on the change! LGTM

PR #148, opened 2024-04-02
Apply search and filters on pressing enter

### LINKED ISSUE #146: Search terms not applied on Enter
When I press Enter while in the search field, I'd expect the filter to be applied

omar-selo:
Resolves #146 

Basically, when a user is focused on a search bar and clicks enter, then the search query and filters selected are applied immediately.

### DIFF
--- a/frontend/lib/ui/page_filters/page_filters.dart
+++ b/frontend/lib/ui/page_filters/page_filters.dart
@@ -29,23 +29,20 @@ class PageFiltersView extends ConsumerWidget {
       child: ListView.separated(
         shrinkWrap: true,
         itemBuilder: (_, i) {
-          if (i == 0) return PageSearchBar(hintText: searchHint);
+          if (i == 0) {
+            return PageSearchBar(
+              hintText: searchHint,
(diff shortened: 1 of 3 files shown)


andrejvelichkovski (approved):
LGTM! Although it is not part of this issue, if it is not a significant change, maybe we can make the Test Execution review pop over submit and close when enter is pressed as well

omar-selo:
> maybe we can make the Test Execution review pop over submit and close when enter is pressed as well

Right now if we press enter a new line is inserted. That sort of makes sense cause the comment can be multi-line. Not sure if changing it to submit would be better.

andrejvelichkovski:
> > maybe we can make the Test Execution review pop over submit and close when enter is pressed as well
> 
> Right now if we press enter a new line is inserted. That sort of makes sense cause the comment can be multi-line. Not sure if changing it to submit would be better.

Oh, I forgot the input box can be multi-line. I think it makes more sense to keep it that way, otherwise we might end up submitting the review while the reviewer still wants to add some comments.

PR #152, opened 2024-04-05
Automatic approval of test executions

omar-selo:
Resolves [RTW-265](https://warthogs.atlassian.net/browse/RTW-265).

Changes:
- Whenever the results of a test execution is submitted to TO from C3, approve that test execution automatically if it passes certain criteria (specified in spec [CR060](https://docs.google.com/document/d/1Nc_aRE8KHO64UkWX8Y2nAzM0ulcuSBi7Pe7f760eC54/edit))
- Split test executions controller into multiple files as it was getting too big
- Use DataGenerator in test executions test cases
- Remove ids passed into `EndTestExecutionRequest` as they aren't necessary and aren't used anywhere
- Updated test results report endpoint to include a few useful fields as [requested](https://chat.canonical.com/canonical/pl/pj5r8n54sjdwug9o31seu7mjuc) 

[RTW-265]: https://warthogs.atlassian.net/browse/RTW-265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


omar-selo:
Sorry for the big diff. Most of it is just splitting test executions controller file into multiple files

backend/test_observer/controllers/test_executions/end_test.py
```diff
@@ -0,0 +1,181 @@
...
+    test_cases = {
+        tr.test_case.name
+        for tr in test_execution.test_results
+        if tr.status != TestResultStatus.SKIPPED
+    }
+
+    return prev_test_cases.issubset(test_cases)
+
+
+def _get_previous_test_execution(
```
andrejvelichkovski, line 80: Could we potentially reuse some of the code from the `get_previous_test_executions_query` defined in `backend/test_observer/controllers/test_executions/logic.py`. I remember implementing the similar functionality for fetching the historic test results for a test execution.

backend/test_observer/controllers/test_executions/end_test.py
```diff
@@ -0,0 +1,181 @@
...
+    if family.name == FamilyName.SNAP.value:
+        query = query.where(Artefact.track == artefact.track)
+
+    if family.name == FamilyName.DEB.value:
+        query = query.where(
+            Artefact.series == artefact.series, Artefact.repo == artefact.repo
+        )
+
+    query = query.order_by(Artefact.id.desc()).limit(1)
+
```
andrejvelichkovski, line 122: I think with our current setup we have NULL values for series/repo for snap artefacts, and similarly for deb artefacts we have NULL values for track fields. Maybe we can simplify this code by just trying to filter based on these three fields, not having separate logic for snaps/debs

backend/test_observer/controllers/test_executions/end_test.py
```diff
@@ -0,0 +1,181 @@
...
+from test_observer.data_access.repository import get_or_create
+from test_observer.data_access.setup import get_db
+
+from .logic import delete_previous_results
+from .models import C3TestResult, C3TestResultStatus, EndTestExecutionRequest
+
+router = APIRouter()
+
+
+@router.put("/end-test")
```
andrejvelichkovski, line 28: We seem to start making a hierarchy of the API endpoints. I remember FastAPI has a support for [tags](https://fastapi.tiangolo.com/tutorial/path-operation-configuration/#tags) which let you split the docs view in separate parts, maybe we can tag these endpoints with `tags=["test-execution"]` to logically separate them in different parts and make the docs a little bit more organized.

backend/test_observer/controllers/test_executions/end_test.py
```diff
@@ -0,0 +1,181 @@
...
+        TestExecutionStatus.FAILED if has_failures else TestExecutionStatus.PASSED
+    )
+
+    prev_test_execution = _get_previous_test_execution(db, test_execution)
+
+    if (
+        prev_test_execution
+        and not has_failures
+        and prev_test_execution.is_approved
+        and _ran_all_previously_run_cases(prev_test_execution, test_execution)
```
andrejvelichkovski, line 50: On one of the previous standups we mentioned making sure that the set of SKIPPED tests hasn't changed, to help us make sure not a lot of test cases started getting skipped instead of passing. Should we add that verification here as well?

backend/test_observer/controllers/test_executions/end_test.py
```diff
@@ -0,0 +1,181 @@
+from fastapi import APIRouter, Depends, HTTPException
```
nadzyah, line 1: Don't forget to add copyright comment

backend/test_observer/controllers/test_executions/__init__.py
```diff
@@ -0,0 +1,9 @@
+from fastapi import APIRouter
```
nadzyah, line 1: Same here

andrejvelichkovski (changes requested):
The code look good to me! Good call for splitting the large files into smaller chunks. I just have one concern about verifying that the set of skipped tests hasn't grown in the current test execution.

All my other comments are subjective about improving the code quality.

backend/test_observer/controllers/test_executions/end_test.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def _parse_c3_test_result_status(status: C3TestResultStatus) -> TestResultStatus:
+    match status:
+        case C3TestResultStatus.PASS:
+            return TestResultStatus.PASSED
+        case C3TestResultStatus.FAIL:
+            return TestResultStatus.FAILED
+        case C3TestResultStatus.SKIP:
+            return TestResultStatus.SKIPPED
```
nadzyah, line 181: Not sure which approach is more effective, but you can just construct a dictionary here like `{c3_result: to_result}` and return `mapped_results[status]`

backend/test_observer/controllers/test_executions/logic.py
```diff
@@ -6,14 +6,12 @@
 from test_observer.data_access.models import (
```
nadzyah, line 6: You forgot about copyright in this file as well

backend/test_observer/controllers/test_executions/patch.py
```diff
@@ -0,0 +1,40 @@
+from fastapi import APIRouter, Depends, HTTPException
```
nadzyah, line 1: And here

backend/tests/controllers/test_executions/test_end_test.py
```diff
@@ -0,0 +1,246 @@
+from fastapi.testclient import TestClient
```
nadzyah, line 1: And here

backend/tests/controllers/test_executions/test_get_test_results.py
```diff
@@ -0,0 +1,71 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Written by:
+#        Omar Selo <omar.selo@canonical.com>
+#        Nadzeya Hutsko <nadzeya.hutsko@canonical.com>
```
nadzyah, line 19: You can remove me here from the authors since you're the only author of this code😅

backend/tests/controllers/test_executions/test_patch.py
```diff
@@ -0,0 +1,59 @@
+import pytest
```
nadzyah, line 1: Copyright here

backend/tests/controllers/test_executions/test_start_test.py
```diff
@@ -0,0 +1,180 @@
+from fastapi.testclient import TestClient
```
nadzyah, line 1: And here

nadzyah (changes requested):
Clean code that is easy to read and understand (as always😊)
Most of the comments are copyright related changes

backend/test_observer/controllers/test_executions/end_test.py
omar-selo: This is partially covered by `_ran_all_previously_run_cases`. In the sense that any tests that have run before (whether passed or failed) have to have run this time too. Together with `not has_failures` ensures that any test that ran before must have passed this time and not skipped. 

But it could be that there are new tests (didn't run last time) that have been skipped. This sounds fine to me, but if you are worried about this, then we can discuss it with others.

backend/test_observer/controllers/test_executions/end_test.py
omar-selo: I thought about this as it is very similar functionality as you mentioned. But it doesn't seem possible to reuse that logic as is, and modifying it to serve both use cases seemed challenging for me (and I worried it would make the code less readable). But if you have an idea of how we can merge the two use cases without over complicating things I would definitely be interested.

backend/test_observer/controllers/test_executions/end_test.py
andrejvelichkovski: Ah I didn't notice the condition`if tr.status != TestResultStatus.SKIPPED` which actually accounts for what I was mentioning. Sorry for this.

In terms of new test cases, I'm not sure what would be the best way to resolve them. In some cases we might want to check explicitly the outcome of the new test cases. I wonder if we shouldn't approve executions where there are any additional test cases compared to previous run, or this wouldn't need any attention of the reviewer?

backend/test_observer/controllers/test_executions/end_test.py
andrejvelichkovski: Wouldn't it be enough to call the `get_previous_test_executions_query` method with `.limit(1)` instead of `.limit(PREVIOUS_TEST_RESULT_COUNT)`. Although that will return just the ID of the test execution, so we would need one extra query to fetch the actual test execution.

This will hurt the performance slightly as we will be making one extra database query and maybe over complicating the logic.

I don't mind either approaches, we can also keep the logic for these two functionalities separate, as they do have some differences as well.

backend/test_observer/controllers/test_executions/end_test.py
omar-selo: Good point

backend/test_observer/controllers/test_executions/end_test.py
andrejvelichkovski: I agree with you, even though we might be able to reuse a little bit of code, we will still need to write a few adapters and methods that will make it even more difficult to understand.

Let's keep it as it is and we can work on uniting the logic if we need to add more layers of complexity on top of it.

backend/test_observer/controllers/test_executions/end_test.py
omar-selo: At least no one brought this up in the spec. I assume it's safe to have new test cases that pass or skip. But we can definitely add this check if we decided it's important later on.

backend/test_observer/controllers/test_executions/end_test.py
omar-selo: the match/case approach works better with mypy. Since the return type is not optional, mypy would complain if we missed a case. Using a dictionary on the other hand would not give us the same guarantees.

andrejvelichkovski (approved):
Thanks for addressing all comments! The code looks good to me

nadzyah (approved):

PR #154, opened 2024-04-09
For non-kernel artefacts, set due date to be today + 10 days

nadzyah:
This PR closes [RTW-266](https://warthogs.atlassian.net/browse/RTW-266)

It sets due_date to be + 10 days from today for non-kernel artefacts, aka artefacts with names that doesn't start with `linux-` and doesn't end with `-kernel`.

The tests were added for the `start-test` endpoint, since right now this is the only place we are creating artefacts from.

[RTW-266]: https://warthogs.atlassian.net/browse/RTW-266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/test_observer/data_access/event_listeners.py
```diff
@@ -0,0 +1,49 @@
...
+        return
+    if target.name.startswith("linux-") or target.name.endswith("-kernel"):
+        # Snap artefacts can also be kernels, we ignore them
+        return
+    if target.stage_id is not None:
+        session = Session(bind=connection)
+
+        stage = session.get(Stage, target.stage_id)
+
+        if stage and stage.family and stage.family.name != FamilyName.DEB.value:
```
omar-selo, line 46: In the future, we could test debs that are not kernels. Assuming all kernels either start with `linux-` or end with `-kernel` then there is no need to require the artefact to be a snap. It should be enough to check the name.

backend/test_observer/data_access/event_listeners.py
```diff
@@ -0,0 +1,49 @@
...
+
+@event.listens_for(Artefact, "before_insert")
+def receive_before_insert(
+    mapper: Mapper, connection: Connection, target: Artefact  # noqa: ARG001
+):
+    """For non-kernel artefacts, set the defaule due date to now() + 10 days"""
+    if target.due_date is not None:
+        # Don't overwrite existing due date
+        return
+    if target.name.startswith("linux-") or target.name.endswith("-kernel"):
```
omar-selo, line 38: It would be great if there was a [hybrid attribute](https://docs.sqlalchemy.org/en/20/orm/extensions/hybrid.html) on the Artefact model that abstracts away the logic of identifying whether an artefact is a kernel. Something like `is_kernel`. If a hybrid attribute proves too challenging to implement, a simple `@attribute` would suffice too.

backend/test_observer/data_access/event_listeners.py
```diff
@@ -0,0 +1,49 @@
...
+from sqlalchemy.engine.base import Connection
+
+from .models import Artefact, Stage
+from .models_enums import FamilyName
+
+
+@event.listens_for(Artefact, "before_insert")
+def receive_before_insert(
+    mapper: Mapper, connection: Connection, target: Artefact  # noqa: ARG001
+):
```
omar-selo, line 33: It might be simpler to use a use a [context sensitive default](https://docs.sqlalchemy.org/en/20/core/defaults.html#context-sensitive-default-functions) instead of an event listener

omar-selo (changes requested):

omar-selo (approved):
Looks great!

PR #157, opened 2024-04-17
Upgrade flutter

omar-selo:
Main reason for this upgrade is to make use of this small improvement https://dart.dev/tools/non-promotion-reasons#property


andrejvelichkovski (changes requested):
Seems like the CI checks are failing because some of the methods we use are deprecated in the new version. Should we replace these with the suggestions?

nadzyah (approved):

PR #158, opened 2024-04-17
Add template ids

omar-selo:
Resolves [CER-2574](https://warthogs.atlassian.net/browse/CER-2574).

Changes:
- Adds template id to test case model
- Populate template id from end test controller
- Expose template id in reports

### DIFF
--- a/backend/test_observer/controllers/test_executions/end_test.py
+++ b/backend/test_observer/controllers/test_executions/end_test.py
@@ -165,7 +165,10 @@ def _store_test_results(
             db,
             TestCase,
             filter_kwargs={"name": r.name},
-            creation_kwargs={"category": r.category},
+            creation_kwargs={
+                "category": r.category,
+                "template_id": r.template_id or "",
+            },
         )
 
         test_result = TestResult(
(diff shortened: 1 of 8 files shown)


backend/test_observer/controllers/test_executions/end_test.py
```diff
@@ -165,7 +165,10 @@ def _store_test_results(
             db,
             TestCase,
             filter_kwargs={"name": r.name},
-            creation_kwargs={"category": r.category},
+            creation_kwargs={
+                "category": r.category,
+                "template_id": r.template_id or "",
+            },
```
andrejvelichkovski, line 171: This will help us make sure that we have template ID for new test cases that we haven't seen till now, but for repeating (already existing tests in our database) it would not change them.

On the C3 side I added an additional check for existing test cases to check if there is template id in the request and the testcase.template_id is null in which case we would make an update to the testcase to set the template id.

Perhaps we can do something similar here?

omar-selo: Ah good point, I completely missed this

PR #159, opened 2024-04-18
Fix template id db migration

omar-selo:
Changes:
- Fix database migration bug introduced in previous [template id PR](https://github.com/canonical/test_observer/pull/158)
- Force charm code to exit if migrations fail to apply. Otherwise the charm `BlockedStatus("Database migration failed")` status gets overwritten by `ActiveStatus()` somewhere else in the code. If it's overwritten then we wouldn't know that migrations failed from juju status, we'd have to manually examine the logs.

### DIFF
--- a/backend/charm/src/charm.py
+++ b/backend/charm/src/charm.py
@@ -113,6 +113,7 @@ class TestObserverBackendCharm(CharmBase):
             logger.error(e.stdout)
             logger.error(e.stderr)
             self.unit.status = BlockedStatus("Database migration failed")
+            raise SystemExit(0)
 
     def _on_database_changed(self, event):
         self._migrate_database()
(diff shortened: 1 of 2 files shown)


andrejvelichkovski (approved):
Good catch! I wonder if something like this might also be happening in the C3 charm with the recent DB migration issue

backend/charm/src/charm.py
```diff
@@ -113,6 +113,7 @@ def _migrate_database(self):
             logger.error(e.stdout)
             logger.error(e.stderr)
             self.unit.status = BlockedStatus("Database migration failed")
+            raise SystemExit(0)
```
nadzyah, line 116: I would like to suggest using `sys.exit(0)` as it’s used in the docs: https://docs.python.org/3/library/exceptions.html#SystemExit

omar-selo: I think both can be used interchangeably. The docs doesn't seem to recommend one over the other. Personally, what I like about  `raise SystemExit(0)` is that it's more clear that it breaks flow and not just a regular function call. But don't really mind either.

PR #171, opened 2024-05-07
Keep reviews after test execution rerun

omar-selo:
First PR towards resolving https://warthogs.atlassian.net/browse/RTW-307

Changes:
- Rerunning a test execution now no longer dismisses review decision or comment
- Following boy scout rule, I've done a bit of code cleaning

### DIFF
--- a/backend/tests/controllers/test_executions/test_start_test.py
+++ b/backend/tests/controllers/test_executions/test_start_test.py
@@ -14,9 +14,13 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+from collections.abc import Callable
 from datetime import date, timedelta
+from typing import Any, TypeAlias
 
+import pytest
 from fastapi.testclient import TestClient
+from httpx import Response
 from sqlalchemy.orm import Session
 
 from test_observer.controllers.test_executions.models import StartTestExecutionRequest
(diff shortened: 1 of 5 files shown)


backend/tests/controllers/test_executions/test_start_test.py
```diff
@@ -29,15 +33,25 @@
...
 
+Execute: TypeAlias = Callable[[dict[str, Any]], Response]
 
-def test_creates_all_data_models(db_session: Session, test_client: TestClient):
-    response = test_client.put(
-        "/v1/test-executions/start-test",
-        json={
+
+@pytest.fixture
+def execute(test_client: TestClient) -> Execute:
```
andrejvelichkovski, line 45: `execute` seems to be a bit too generic for this helper method. Perhaps we can rename it to `start_test_execution` or something else that will indicate that this method makes a `PUT` request to start the test execution?

andrejvelichkovski (approved):
Looks good to me! I didn't know about the boy scout rule, but looks like a great way to keep remembering to improve existing code.

backend/tests/controllers/test_executions/test_start_test.py
omar-selo: I opted for the generic name because this whole file is called `test_start_test.py` so by definition execute would just start the test

PR #172, opened 2024-05-07
Fix bug with promote script

omar-selo:
Found a bug with promote script. The helper method `queries.artefact_architectures` filters by `ArtefactBuild.id == artefact_id` where it should've filtered by `ArtefactBuild.artefact_id == artefact_id`. This meant that in some cases promote script couldn't promote debs.

This is a particularly nasty bug because the tests pass. But if you ran pytest on just `promote.py` file then the test would fail. Which begs the question, why? The answer turns out to be because database transaction rollback doesn't reset the auto increment value of a primary key. Hence in some cases it could just happen that `ArtefactBuild.id == artefact_id`. In other cases it would not be. That all depends on what tests ran before this particular failing test. (Learning point, database transaction rollback doesn't rollback the incrementation of primary keys)

### DIFF
--- a/backend/test_observer/controllers/promoter/promoter.py
+++ b/backend/test_observer/controllers/promoter/promoter.py
@@ -187,9 +187,7 @@ def run_deb_promoter(session: Session, artefact: Artefact) -> None:
     assert series is not None, f"Series is not set for the artefact {artefact.id}"
     assert repo is not None, f"Repo is not set for the artefact {artefact.id}"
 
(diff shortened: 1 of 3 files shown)


andrejvelichkovski (approved):
Great investigation on this! The fact that auto-increment keys don't rollback is new for me as well, I wasn't expecting them to work that way, motivated that transactions either fully work or don't make any changes to the database

PR #173, opened 2024-05-07
Rerun artefact test executions

omar-selo:
This is the backend side of https://warthogs.atlassian.net/browse/RTW-307

Changes:
- Adds a new endpoint `POST /v1/artefacts/<artefact-id>/reruns` that allows user to rerun all test executions belonging to an artefact
- Adds some useful filters to the above endpoint

Note that CI will fail until we merge and rebase https://github.com/canonical/test_observer/pull/172

### DIFF
--- a/backend/test_observer/controllers/artefacts/artefacts.py
+++ b/backend/test_observer/controllers/artefacts/artefacts.py
@@ -164,7 +164,7 @@ def rerun_artefact_test_executions(
             test_executions = (te for te in test_executions if te.status == status)
         if (decision := request.test_execution_review_decision) is not None:
             test_executions = (
-                te for te in test_executions if te.review_decision == decision
+                te for te in test_executions if set(te.review_decision) == decision
             )
 
     for te in test_executions:
--- a/backend/test_observer/controllers/artefacts/artefacts.py
+++ b/backend/test_observer/controllers/artefacts/artefacts.py
@@ -21,9 +21,14 @@ from fastapi import APIRouter, Depends, HTTPException
 from sqlalchemy.orm import Session, joinedload
 
 from test_observer.data_access import queries
-from test_observer.data_access.models import Artefact, ArtefactBuild, TestExecution
+from test_observer.data_access.models import (
+    Artefact,
+    ArtefactBuild,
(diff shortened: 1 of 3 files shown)


backend/test_observer/controllers/artefacts/models.py
```diff
@@ -93,3 +93,8 @@ class ArtefactBuildDTO(BaseModel):
 
 class ArtefactPatch(BaseModel):
     status: ArtefactStatus
+
+
+class RerunArtefactTestExecutionsRequest(BaseModel):
+    test_execution_status: TestExecutionStatus | None = None
+    test_execution_review_decision: list[TestExecutionReviewDecision] | None = None
```
andrejvelichkovski, line 100: The `review_decision` is defined as a set in the `TestExecutionDTO`, although here it wouldn't really change a lot, if you want we can make it a set

backend/test_observer/controllers/artefacts/artefacts.py
```diff
@@ -137,3 +146,26 @@ def get_artefact_builds(artefact_id: int, db: Session = Depends(get_db)):
...
+        queries.latest_artefact_builds.where(ArtefactBuild.artefact_id == artefact.id)
+    )
+    test_executions = (te for ab in latest_builds for te in ab.test_executions)
+
+    if request:
+        if status := request.test_execution_status:
+            test_executions = (te for te in test_executions if te.status == status)
+        if (decision := request.test_execution_review_decision) is not None:
+            test_executions = (
+                te for te in test_executions if te.review_decision == decision
```
andrejvelichkovski, line 167: If we don't define the review decision as a set maybe here we should do set comparison instead of list as the order doesn't really matter.

andrejvelichkovski (dismissed):
Overall looks good to me, just one concern about the review decision filter

backend/test_observer/controllers/artefacts/artefacts.py
omar-selo: ah very good point

andrejvelichkovski (approved):
LGTM! Thanks for addressing the comment!

PR #179, opened 2024-05-09
Rerun filtered environments button

omar-selo:
Resolves https://warthogs.atlassian.net/browse/RTW-307

Changes:
- Modify `POST /v1/test-executions/reruns` endpoint to take a set of test execution ids to rerun instead of just one
- Remove `POST /v1/artefacts/<artefact-id>/reruns` endpoint as I used the endpoint above instead. The reason is that this endpoint will have to filter exactly as the frontend does. This could be prone to errors, so instead I'd rather rerun exactly what the frontend user can see
- Add the button `RerunFilteredEnvironments` you can see how it works in the video below
- Replace `filteredTestExecutionIds` provider with `filteredTestExecutions`

[Screencast from 2024-05-09 13-34-33.webm](https://github.com/canonical/test_observer/assets/4087200/33ab6ab3-016e-4490-9bf8-c9e7d20f2083)


backend/test_observer/controllers/artefacts/models.py
```diff
@@ -98,3 +98,4 @@ class ArtefactPatch(BaseModel):
 class RerunArtefactTestExecutionsRequest(BaseModel):
     test_execution_status: TestExecutionStatus | None = None
     test_execution_review_decision: set[TestExecutionReviewDecision] | None = None
+    test_execution_environment_contains: str | None = None
```
andrejvelichkovski, line 101: Should we delete this Request model now

backend/test_observer/controllers/test_executions/reruns.py
```diff
@@ -18,10 +18,15 @@
...
+def create_rerun_requests(request: RerunRequest, db: Session = Depends(get_db)):
+    for test_execution_id in request.test_execution_ids:
+        _create_rerun_request(test_execution_id, db)
+
+
+def _create_rerun_request(test_execution_id: int, db: Session) -> None:
+    te = db.get(TestExecution, test_execution_id)
     if not te:
-        msg = f"No test execution with id {request.test_execution_id} found"
+        msg = f"No test execution with id {test_execution_id} found"
```
andrejvelichkovski, line 29: This way of checking whether the Test Execution ID is valid might lead us to the API behaving somewhat inconsistent. For example, if only one of the test execution IDs provided is invalid, then all test executions appearing before it would have been rerun, while the ones after it wont.

Maybe a better way we can handle this is to firstly filter just test executions whose id is in the set, and then create `TestExecutionRerunRequest` just for those? Or maybe just ignore the ones that have invalid ID, instead of raising the `HTTPException`

andrejvelichkovski (changes requested):
Overall looks good to me, but I have two small concerns

backend/test_observer/controllers/test_executions/reruns.py
omar-selo: Good point. I opted to rerun everything that can be rerun. Then return a list of what was rerun. If nothing gets rerun then I return a status code of 404. If everything gets rerun then a status code of 200. And if some rerun and others don't then I return 207.

andrejvelichkovski (approved):
Thanks for addressing the issues! Looks good to me now!

PR #183, opened 2024-06-05
Fix delete rerun doesn't actually work

omar-selo:
Our testing infrastructure sadly can't detect whether a sqlalchemy session has been committed or not. This led before to bugs leaking to production. And it has done so now too. Turns out this endpoint doesn't work. Which is why [reruns can still keep rerunning multiple times](https://chat.canonical.com/canonical/pl/rfrxxqpg43dttqox957cwtscmw).


andrejvelichkovski (approved):
LGTM!

PR #184, opened 2024-06-07
Add status update endpoint to receive updates from Testflinger

val500:
Changes:
- Adds a new endpoint in POST /v1/test-executions/{test-execution-id}/status_update
- Creates two new tables for storing status updates and their corresponding list of test events


backend/test_observer/data_access/models.py
```diff
@@ -402,3 +406,41 @@ def __repr__(self) -> str:
...
+    A table to represent status updates for the test runs
+    """
+
+    __tablename__ = "test_status_update"
+
+    agent_id: Mapped[str]
+    job_queue: Mapped[str]
+    events: Mapped[list["TestEvent"]] = relationship(
+        back_populates="test_status_update", cascade="all, delete"
+    )
```
omar-selo, line 422: it would be useful to have the events relationship order by timestamp by default

backend/test_observer/controllers/test_executions/status_update.py
```diff
@@ -0,0 +1,63 @@
...
+from test_observer.data_access.setup import get_db
+
+from .logic import delete_previous_status_update
+from .models import StatusUpdateRequest
+
+router = APIRouter()
+
+
+@router.post("/{id}/status_update")
+def status_update(id: int, request: StatusUpdateRequest, db: Session = Depends(get_db)):
```
omar-selo, line 35: PUT might be better here as it shows that the endpoint is idempotent and replaces rather than creates

backend/test_observer/controllers/test_executions/models.py
```diff
@@ -172,3 +174,15 @@ class PendingRerun(BaseModel):
 
 class DeleteReruns(BaseModel):
     test_execution_ids: set[int]
+
+
+class JobEvent(BaseModel):
+    event_name: TestEventEnum
+    timestamp: datetime
+    detail_msg: str
```
omar-selo, line 182: I would use `TestEvent` to be more consistent

omar-selo (changes requested):

backend/test_observer/data_access/models_enums.py
```diff
@@ -60,3 +60,25 @@ class TestResultStatus(str, Enum):
...
+    ENDED_FIRMWARE_UPDATE = "ended_firmware_update"
+    ENDED_TEST = "ended_test"
+    ENDED_ALLOCATE = "ended_allocate"
+    ENDED_RESERVE = "ended_reserve"
+    ENDED_CLEANUP = "ended_cleanup"
+    CANCELLED = "cancelled"
+    GLOBAL_TIMEOUT = "global_timeout"
+    OUTPUT_TIMEOUT = "output_timeout"
+    RECOVERY_FAILED = "recovery_failed"
+    FAILED = "failed"
```
omar-selo, line 84: TO should ideally not know the events by name as knowing them would couple it to TF. So it would be better if there is no enum

backend/test_observer/data_access/models_enums.py
```diff
@@ -35,6 +35,7 @@ class TestExecutionStatus(str, Enum):
     PASSED = "PASSED"
     FAILED = "FAILED"
     NOT_TESTED = "NOT_TESTED"
+    ENDED = "ENDED"
```
nadzyah, line 38: Aren't passed and failed tests also ended?

val500: Ended here means tests that have completed but do not have test results(eg. provisioning failure).

backend/test_observer/controllers/test_executions/models.py
```diff
@@ -172,3 +173,13 @@ class PendingRerun(BaseModel):
...
+
+
+class TestEvent(BaseModel):
+    event_name: str
+    timestamp: datetime
+    detail: str
+
+
+class StatusUpdateRequest(BaseModel):
+    events: list[TestEvent]
```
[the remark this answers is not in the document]
val500: The way we're doing that now is by adding a JOB_START event and inserting the resource_url into the detail field of this event.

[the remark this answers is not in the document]
omar-selo: Ooh ok, can we then add it to the test execution database object? via the Testflinger parser

omar-selo:
@val500 don't forget to attach a video of the frontend UI as otherwise it's difficult for me to imagine how it'll look like

val500:
> @val500 don't forget to attach a video of the frontend UI as otherwise it's difficult for me to imagine how it'll look like
 Here's the video:
[test_observer_event_log.webm](https://github.com/user-attachments/assets/40348701-f8ee-4787-92c8-c9e03421fa91)

backend/test_observer/data_access/models_enums.py
plars: Wouldn't that also be considered failed from the standpoint of this run if provisioning fails?

val500: It could be considered a failure, but IIRC, we had a meeting and decided to use a different status/icon for tests that don't have results.

omar-selo: Yes it makes sense to have `FAILED` means that there were some tests failing, and `ENDED` meaning that testing didn't complete. Perhaps it's not the best verb to use, as ended doesn't imply that something bad happened. In fact passed or failed both did end. We could call it `ENDED_PREMATURELY`

omar-selo:
> > @val500 don't forget to attach a video of the frontend UI as otherwise it's difficult for me to imagine how it'll look like
> > Here's the video:
> > [test_observer_event_log.webm](https://github.com/user-attachments/assets/40348701-f8ee-4787-92c8-c9e03421fa91)

Nice I like the table. Just the left padding seems a bit off for some test executions/environments

val500:
test_observer_event_log
Updated artefact view - spacing is fixed and the latest event is shown next to each execution as described in the Jira ticket.

omar-selo (approved):
Looks good to me, thanks for being patient and following up with our concerns!

PR #188, opened 2024-07-08
RTW-321: Deprecate two enum choices

andrejvelichkovski:
## Description

This change deprecates the two following options from the Test Execution Review Decision:
* Unstable Physical Infrastructure
* Faulty Hardware

The deprecation is only done on the frontend. To make sure everything is backwards compatible, we will still show the cases where these are selected, but they will be disabled and their selection wont be modifiable.

## Resolved issues

RTW-321

## Documentation

No documentation changes involved.

## Web service API changes

No API changes involved.

## Tests

The testing was performed manually. Firstly, we observe the case where one of the options was already selected:

Screenshot from 2024-07-08 10-42-21

We then observe that the choices are not rendered for new test executions

Screenshot from 2024-07-08 10-42-37

### DIFF
--- a/frontend/lib/models/test_execution.dart
+++ b/frontend/lib/models/test_execution.dart
@@ -127,6 +127,19 @@ enum TestExecutionReviewDecision {
     }
   }
 
+  bool get isDeprecated {
+    switch (this) {
+      case approvedFaultyHardware:
+      case approvedUnstablePhysicalInfra:
+        return true;
+      case rejected:
+      case approvedInconsistentTest:
+      case approvedCustomerPrerequisiteFail:
+      case approvedAllTestsPass:
+        return false;
+    }
+  }
+
   String toJson() {
     switch (this) {
       case rejected:
--- a/frontend/lib/models/test_execution.dart
+++ b/frontend/lib/models/test_execution.dart
@@ -132,10 +132,7 @@ enum TestExecutionReviewDecision {
       case approvedFaultyHardware:
       case approvedUnstablePhysicalInfra:
         return true;
-      case rejected:
-      case approvedInconsistentTest:
-      case approvedCustomerPrerequisiteFail:
-      case approvedAllTestsPass:
+      default:
         return false;
     }
   }
--- a/frontend/lib/ui/artefact_page/test_execution_pop_over.dart
+++ b/frontend/lib/ui/artefact_page/test_execution_pop_over.dart
@@ -36,10 +36,12 @@ class TestExecutionPopOverState extends ConsumerState<TestExecutionPopOver> {
   Function(bool?)? getOnChangedCheckboxListTileFunction(
     TestExecutionReviewDecision testExecutionReviewDecision,
   ) {
-    if ((testExecutionReviewDecision == TestExecutionReviewDecision.rejected &&
-            (_canReject)) ||
-        (testExecutionReviewDecision != TestExecutionReviewDecision.rejected &&
-            _canApprove)) {
+    if (!testExecutionReviewDecision.isDeprecated &&
+        ((testExecutionReviewDecision == TestExecutionReviewDecision.rejected &&
+                _canReject) ||
+            (testExecutionReviewDecision !=
+                    TestExecutionReviewDecision.rejected &&
+                _canApprove))) {
       return (bool? value) {
         setState(() {
           if (reviewDecisions.contains(testExecutionReviewDecision)) {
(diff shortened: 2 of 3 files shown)


frontend/lib/ui/artefact_page/test_execution_review.dart
```diff
@@ -49,7 +49,7 @@ class TestExecutionReviewButton extends StatelessWidget {
           ),
           direction: PopoverDirection.bottom,
           width: 500,
-          height: 500,
+          height: 400,
```
andrejvelichkovski, line 52: The PopOver library doesn't support automatic sizing, so I have adapted the size to fit with less choices.

frontend/lib/models/test_execution.dart
```diff
@@ -127,6 +127,16 @@ enum TestExecutionReviewDecision {
     }
   }
 
+  bool get isDeprecated {
+    switch (this) {
+      case approvedFaultyHardware:
+      case approvedUnstablePhysicalInfra:
+        return true;
```
nadzyah, line 134: sorry if I don't know how dart works, but shouldn't it look like this:

```suggestion
      case approvedFaultyHardware:
        return true;
      case approvedUnstablePhysicalInfra:
        return true;
```

andrejvelichkovski: According to the [Dart docs](https://dart.dev/language/branches#switch-statements) the empty case falls through, so in this case the logic for `approvedFaultyHardware` continues through and executes the same case as `approvedUnstablePhysicalInfra`.

frontend/lib/ui/artefact_page/test_execution_pop_over.dart
```diff
@@ -53,6 +55,11 @@ class TestExecutionPopOverState extends ConsumerState<TestExecutionPopOver> {
     return null;
   }
 
+  bool shouldDisplayDecision(TestExecutionReviewDecision decision) {
+    return !decision.isDeprecated ||
+        (decision.isDeprecated && reviewDecisions.contains(decision));
```
omar-selo, line 60:
```suggestion
    return !decision.isDeprecated || reviewDecisions.contains(decision);
```

frontend/lib/ui/artefact_page/test_execution_pop_over.dart
```diff
@@ -36,10 +36,12 @@ class TestExecutionPopOverState extends ConsumerState<TestExecutionPopOver> {
...
-    if ((testExecutionReviewDecision == TestExecutionReviewDecision.rejected &&
-            (_canReject)) ||
-        (testExecutionReviewDecision != TestExecutionReviewDecision.rejected &&
-            _canApprove)) {
+    if (!testExecutionReviewDecision.isDeprecated &&
+        ((testExecutionReviewDecision == TestExecutionReviewDecision.rejected &&
+                _canReject) ||
+            (testExecutionReviewDecision !=
+                    TestExecutionReviewDecision.rejected &&
+                _canApprove))) {
```
omar-selo, line 44: can you simplify or make this if statement easier to understand somehow? one suggestion is to just create a variable whose name explains `((testExecutionReviewDecision == TestExecutionReviewDecision.rejected &&
                _canReject) ||
            (testExecutionReviewDecision !=
                    TestExecutionReviewDecision.rejected &&
                _canApprove))`

frontend/lib/ui/artefact_page/test_execution_review.dart
omar-selo: What happens if there were deprecated options selected? would the user have to scroll in the popover? can't say I like that

omar-selo (changes requested):

frontend/lib/ui/artefact_page/test_execution_review.dart
andrejvelichkovski: Yeah, the user would have to scroll a bit. I couldn't find a way to add some dynamic sizing to this widget. Perhaps the other alternative would be to count those and add some sizing based on how many we should show, but not sure how nice that would look in code.

omar-selo: Yeah the alternative is not great either. This package isn't very flexible. I guess we can keep the scroll

frontend/lib/ui/artefact_page/test_execution_pop_over.dart
andrejvelichkovski: I moved the second part in the new variable as suggested, and added a comment that explains what is happening here.

omar-selo (approved):
LGTM!

PR #190, opened 2024-07-17
Add artefact due date filters

omar-selo:
## Description

Adds a new artefacts filter based on due date.

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-324

## Documentation

I don't think we have any documentation on filters.

## Web service API changes

None

## Tests

See below video:
[Screencast from 2024-07-17 15-02-12.webm](https://github.com/user-attachments/assets/7266ecd1-6cfd-492f-9bba-020ef45c1eea)

### DIFF
--- a/frontend/lib/models/filters.dart
+++ b/frontend/lib/models/filters.dart
@@ -91,6 +91,22 @@ final emptyArtefactFilters = Filters<Artefact>(
       name: 'Status',
       extractOption: (artefact) => artefact.status.name,
     ),
+    Filter<Artefact>(
+      name: 'Due date',
+      extractOption: (artefact) {
+        final now = DateTime.now();
+        final dueDate = artefact.dueDate;
+
+        if (dueDate == null) return 'No due date';
+        if (dueDate.isBefore(now)) return 'Overdue';
+
+        final daysDueIn = now.difference(dueDate).inDays;
+
+        if (daysDueIn >= 14) return 'Later than next week';
+        if (daysDueIn >= 7) return 'Next week';
+        return 'This week';
+      },
+    ),
   ],
 );
 


andrejvelichkovski (dismissed):
Wow, really nice and clean implementation. I guess the background preparation we did for other filters helped us implement this one quite easily.

My only comment was about the "This week" option, as it is not really this week, but rather in a weeks time from now. So maybe we can rename the options to:
* This week - In a week
* Next week - In more than a week
* Later than next week - In more than two weeks

Still, I'm not really sure what would be the best choice myself. Perhaps @jocave as a native-speaker should help us figure out the best wording for the choices.

jocave:
> Wow, really nice and clean implementation. I guess the background preparation we did for other filters helped us implement this one quite easily.
> 
> My only comment was about the "This week" option, as it is not really this week, but rather in a weeks time from now. So maybe we can rename the options to:
> 
> * This week - In a week
> * Next week - In more than a week
> * Later than next week - In more than two weeks
> 
> Still, I'm not really sure what would be the best choice myself. Perhaps @jocave as a native-speaker should help us figure out the best wording for the choices.

within a week
within two weeks
more than two weeks

?

omar-selo:
> within a week
> within two weeks
> more than two weeks

Sounds clear and concise

omar-selo:
I suggest we go with "Within a week" and "More than a week" and drop "More than two weeks" because I don't think we can actually see that at all. IIRC no artefacts will have more than two weeks deadline

andrejvelichkovski:
> I suggest we go with "Within a week" and "More than a week" and drop "More than two weeks" because I don't think we can actually see that at all. IIRC no artefacts will have more than two weeks deadline

Seems reasonable for now, we can follow-up if there are any concerns from others

andrejvelichkovski (approved):
LGTM!

PR #195, opened 2024-07-30
Add rerun test executions helper script

omar-selo:
## Description

This is a script helper script for reviewers. Often when reviewing an artefact, we'd want to rerun all test executions that failed a particular test or group of tests. This handy script does just that for us as otherwise we'd have to manually go through all of the test executions in the UI which is time consuming and annoying.

```
$ python scripts/rerun_failing_test_executions.py -h
usage: rerun_failing_test_executions.py [-h] artefact_id test_case_regex

Reruns test executions of an artefact that failed particular test cases matched by the passed regex

positional arguments:
  artefact_id
  test_case_regex

options:
  -h, --help       show this help message and exit

Example:

  python rerun_failing_test_executions.py 47906 '.*wireless.*'
  python rerun_failing_test_executions.py 47906 '.*suspend.*'
```

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-337

## Documentation

None.

## Web service API changes

None.

## Tests

This is a throwaway script (ideally this sort of functionality will be available through the UI eventually). But I've tried it myself a few times.

### DIFF
--- a/backend/scripts/rerun_failing_test_executions.py
+++ b/backend/scripts/rerun_failing_test_executions.py
@@ -1,12 +1,16 @@
 # ruff: noqa: T201
 
 import argparse
+import functools
 import re
 from os import environ
 
-import requests
+from requests import Session
 
 TO_API_URL = environ.get("TO_API_URL", "https://test-observer-api.canonical.com")
+
+requests = Session()
+requests.request = functools.partial(requests.request, timeout=5)  # type: ignore
 
 
 def main(artefact_id: int, test_case_regex: str, reversed: bool):
@@ -56,9 +60,9 @@ def main(artefact_id: int, test_case_regex: str, reversed: bool):
     should_rerun = (
         input(
             f"Will rerun {len(test_execution_ids_to_rerun)}"
-            " test executions is that ok? (yes/no) "
+            " test executions is that ok? (y/N) "
         )
-        == "yes"
+        == "y"
     )
 
     if should_rerun:
--- a/backend/scripts/rerun_failing_test_executions.py
+++ b/backend/scripts/rerun_failing_test_executions.py
@@ -10,7 +10,7 @@ from requests import Session
 TO_API_URL = environ.get("TO_API_URL", "https://test-observer-api.canonical.com")
 
 requests = Session()
-requests.request = functools.partial(requests.request, timeout=5)  # type: ignore
+requests.request = functools.partial(requests.request, timeout=30)  # type: ignore
 
 
 def main(artefact_id: int, test_case_regex: str, reversed: bool):
--- a/backend/scripts/rerun_failing_test_executions.py
+++ b/backend/scripts/rerun_failing_test_executions.py
@@ -2,10 +2,11 @@
 
 import argparse
 import re
+from os import environ
 
 import requests
 
-TO_API_URL = "https://test-observer-api.canonical.com"
+TO_API_URL = environ.get("TO_API_URL", "https://test-observer-api.canonical.com")
 
 
 def main(artefact_id: int, test_case_regex: str, reversed: bool):
@@ -15,14 +16,15 @@ def main(artefact_id: int, test_case_regex: str, reversed: bool):
 
     test_case_matcher = re.compile(test_case_regex)
 
-    relevant_test_executions = [
-        te
-        for ab in artefact_builds
-        for te in ab["test_executions"]
-        if te["review_decision"] == []
-        and not te["is_rerun_requested"]
-        and te["status"] == "FAILED"
-    ]
+    relevant_test_executions = []
+    for ab in artefact_builds:
+        for te in ab["test_executions"]:
+            if (
+                te["review_decision"] == []
+                and not te["is_rerun_requested"]
+                and te["status"] == "FAILED"
+            ):
+                relevant_test_executions.append(te)
 
     test_execution_ids_to_rerun = []
     for te in relevant_test_executions:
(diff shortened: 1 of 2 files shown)


backend/scripts/rerun_failing_test_executions.py
```diff
@@ -0,0 +1,87 @@
...
+
+TO_API_URL = "https://test-observer-api.canonical.com"
+
+
+def main(artefact_id: int, test_case_regex: str, reversed: bool):
+    artefact_builds = requests.get(
+        f"{TO_API_URL}/v1/artefacts/{artefact_id}/builds"
+    ).json()
+
+    test_case_matcher = re.compile(test_case_regex)
```
nadzyah, line 16: What if this call fails? Should we handle an exception here?

backend/scripts/rerun_failing_test_executions.py
```diff
@@ -0,0 +1,87 @@
...
+
+    relevant_test_executions = [
+        te
+        for ab in artefact_builds
+        for te in ab["test_executions"]
+        if te["review_decision"] == []
+        and not te["is_rerun_requested"]
+        and te["status"] == "FAILED"
+    ]
```
nadzyah, line 25: I guess it would be more readable to rewrite it as a normal loop

backend/scripts/rerun_failing_test_executions.py
```diff
@@ -0,0 +1,87 @@
...
+            " test executions is that ok? (Y/n) "
+        )
+        == "Y"
+    )
+
+    if should_rerun:
+        requests.post(
+            f"{TO_API_URL}/v1/test-executions/reruns",
+            json={"test_execution_ids": test_execution_ids_to_rerun},
+        )
```
nadzyah, line 66: You can also specify a timeout here

backend/scripts/rerun_failing_test_executions.py
```diff
@@ -0,0 +1,87 @@
...
+    test_execution_ids_to_rerun = []
+    for te in relevant_test_executions:
+        test_results = requests.get(
+            f"{TO_API_URL}/v1/test-executions/{te['id']}/test-results"
+        ).json()
+
+        matching_failed_tests = (
+            tr
+            for tr in test_results
+            if tr["status"] == "FAILED" and test_case_matcher.match(tr["name"])
```
nadzyah, line 36: Probably you can introduce a small pydantic model at the top of the script instead of dealing with raw json

backend/scripts/rerun_failing_test_executions.py
```diff
@@ -0,0 +1,87 @@
...
+                f"will rerun {te['environment']['name']}"
+                f" as no failing matches found and reversed option is set"
+            )
+
+    should_rerun = (
+        input(
+            f"Will rerun {len(test_execution_ids_to_rerun)}"
+            " test executions is that ok? (Y/n) "
+        )
+        == "Y"
```
nadzyah, line 59: “y” should also be an acceptable value, and a simple “enter”
When you specify `Y/n`, the uppercase choice means a default option (probably you already knew that). 

If you want to deal with exact value, it would be more user-friendly to specify it in the following way:

```suggestion
            " test executions is that ok? (yes/no) "
        )
        == "yes"
```

backend/scripts/rerun_failing_test_executions.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+
+    if should_rerun:
+        requests.post(
+            f"{TO_API_URL}/v1/test-executions/reruns",
+            json={"test_execution_ids": test_execution_ids_to_rerun},
+        )
+        print("Submitted rerun requests")
+
```
nadzyah, line 68:
```suggestion
    else:
        print(”Aborting the script”)
```

backend/scripts/rerun_failing_test_executions.py
omar-selo: By the call do you mean the api call or `re.compile` call? Failing the former would most likely a network error (not sure we should handle that). I'm not sure if we can fail the latter based on my understanding of the [docs](https://docs.python.org/3.10/library/re.html#re.compile).

backend/scripts/rerun_failing_test_executions.py
omar-selo: This is meant to be a small script that we should get rid of later on (as everything should be through the api or frontend). I don't think it's necessary to have this added complexity

backend/scripts/rerun_failing_test_executions.py
omar-selo: Oh I didn't know about the uppercase convention

backend/scripts/rerun_failing_test_executions.py
omar-selo: This endpoint doesn't actually wait for the reruns to finish

nadzyah: I mean the request timeout

backend/scripts/rerun_failing_test_executions.py
```diff
@@ -0,0 +1,96 @@
...
+                f"will rerun {te['environment']['name']}"
+                f" as no failing matches found and reversed option is set"
+            )
+
+    should_rerun = (
+        input(
+            f"Will rerun {len(test_execution_ids_to_rerun)}"
+            " test executions is that ok? (y/N) "
+        )
+        == "y"
```
nadzyah, line 65: what if it's `Y`?

nadzyah (dismissed):
Thanks for your work on this!

nadzyah (approved):

PR #197, opened 2024-08-01
HOTFIX: Allow more simultaneous connections to the database

omar-selo:
## Description

API is struggling right now on production due to not enough connections available. This increases the amount. Postgresql currently can handle 100 connections concurrently. But I've set it to 20+30=50 in api server so that in the future we can add another api replica. I hope this is enough.


andrejvelichkovski (approved):
LGTM

PR #198, opened 2024-08-01
Further increase db connection pool

omar-selo:
## Description

Previous increase in connection pool size wasn't enough. This PR further increases it to almost the maximum. Ofc this is not the solution, but it's a possible temporary fix until we get things under control.


andrejvelichkovski (approved):
Let's try if this improves the state

PR #199, opened 2024-08-01
temporarily disable prefetch requests made by frontend to reduce api …

omar-selo:
## Description

This temporary hotfix is in case bumping the number of connections allowed to the db isn't enough. It disables prefetching of test results and test events to reduce the load on API.


andrejvelichkovski (approved):
LGTM!

PR #204, opened 2024-08-09
Fix test events not being loaded

omar-selo:
## Description

When I disabled pre-loading test results and events turns out I accidentally disabled loading test events completely. This fixes that issue.


val500 (approved):
LGTM

PR #206, opened 2024-08-19
Add artefacts sorting

omar-selo:
## Description

Add capability to sort artefacts by different columns from the artefacts list view.

## Resolved issues

https://warthogs.atlassian.net/browse/RTW-219

## Tests

[Screencast from 2024-08-19 14-21-47.webm](https://github.com/user-attachments/assets/7da88a96-2a9c-4ce6-9d2b-659c57699566)

### DIFF
--- a/frontend/lib/providers/filtered_family_artefacts.dart
+++ b/frontend/lib/providers/filtered_family_artefacts.dart
@@ -19,7 +19,8 @@ LinkedHashMap<int, Artefact> filteredFamilyArtefacts(
   final artefacts = ref.watch(familyArtefactsProvider(family)).requireValue;
   final filters =
       emptyArtefactFilters.copyWithQueryParams(pageUri.queryParametersAll);
-  final searchValue = pageUri.queryParameters['q'] ?? '';
+  final searchValue =
+      pageUri.queryParameters[CommonQueryParameters.searchQuery] ?? '';
 
   final filteredArtefacts = artefacts.values
       .filter(
--- a/frontend/lib/providers/filtered_family_artefacts.dart
+++ b/frontend/lib/providers/filtered_family_artefacts.dart
@@ -30,8 +30,9 @@ LinkedHashMap<int, Artefact> filteredFamilyArtefacts(
       )
       .toList();
 
-  final sortBy = pageUri.queryParameters['sortBy'];
-  final sortDirection = pageUri.queryParameters['direction'];
+  final sortBy = pageUri.queryParameters[CommonQueryParameters.sortBy];
+  final sortDirection =
+      pageUri.queryParameters[CommonQueryParameters.sortDirection];
 
   int Function(Artefact, Artefact)? compare;
   switch (sortBy) {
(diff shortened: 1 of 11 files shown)


frontend/lib/providers/filtered_family_artefacts.dart
```diff
@@ -1,28 +1,42 @@
...
 import '../models/filters.dart';
 import '../routing.dart';
+import '../utils/artefact_sorting.dart';
 import 'family_artefacts.dart';
 
 part 'filtered_family_artefacts.g.dart';
 
 @riverpod
-Map<int, Artefact> filteredFamilyArtefacts(
+LinkedHashMap<int, Artefact> filteredFamilyArtefacts(
```
andrejvelichkovski, line 15: TIL: `LinkedHashMap` is the sorted version of a `Map`, really useful in such scenario!

frontend/lib/ui/dashboard/dashboard_body/artefacts_list_view/headers.dart
```diff
@@ -11,21 +11,61 @@ class _Headers extends StatelessWidget {
...
+    Uri pageUri,
+    ColumnMetadata columnData,
+  ) {
+    final queryParameters = pageUri.queryParameters;
+    final sortBy = columnData.queryParam.name;
+    final direction = queryParameters[CommonQueryParameters.sortBy] == sortBy &&
+            queryParameters[CommonQueryParameters.sortDirection] ==
+                SortDirection.asc.name
+        ? SortDirection.desc.name
+        : SortDirection.asc.name;
```
andrejvelichkovski, line 63: Can we simplify the logic here a bit, it seems like we are trying to handle the following: If the header we are pressing is different than the currently chosen one use ascending order, otherwise use descending sort, but it seems a bit difficult to understand written all in one line?

frontend/lib/utils/artefact_sorting.dart
```diff
@@ -0,0 +1,89 @@
...
+    case ArtefactSortingQuery.series:
+      compare = (a1, a2) => a1.series.compareTo(a2.series);
+      break;
+    case ArtefactSortingQuery.repo:
+      compare = (a1, a2) => a1.repo.compareTo(a2.repo);
+      break;
+    case ArtefactSortingQuery.pocket:
+      compare = (a1, a2) => a1.stage.name.compareTo(a2.stage.name);
+      break;
+  }
```
andrejvelichkovski, line 82: Maybe we can move this block to a separate helper `_getArtefactCompareFunction(ArtefactSortingQuery sortByParsed)` which will define and return such function, to simplify the logic and shorten this function?

andrejvelichkovski (dismissed):
Looks really good! Excited to see this land and make the list view more useful! I just left two small style comments.

frontend/lib/ui/dashboard/dashboard_body/artefacts_list_view/headers.dart
omar-selo: Thanks for letting me know that this was difficult to read. You are half right. We only set descending order if both the chosen header is the same as the current one, and the existing direction is ascending. I've updated the code to make this clearer

andrejvelichkovski (approved):
Thanks for addressing the comments!

PR #207, opened 2024-08-29
Add test case issue reporting

omar-selo:
## Description

Adds model and endpoints to support reporting test case issues.

## Resolved issues

Backend part of https://warthogs.atlassian.net/browse/RTW-355

## Web service API changes

New endpoints:
- POST /v1/test-cases/reported-issues to create issues
- GET /v1/test-cases/reported-issues?template_id&case_name to get issues (optionally ones belonging to particular template_id or a particular case_name)
- PUT /v1/test-cases/reported-issues/id to update an issue
- DELETE /v1/test-cases/reported-issues/id to delete an issue

## Tests

Added automated tests.


backend/test_observer/controllers/test_cases/__init__.py
andrejvelichkovski, line 1: Do we need this one, it seems like we are defining `APIRouter` in `test_cases/reported_issues.py` anyway.

backend/tests/controllers/test_cases/test_reported_issues.py
```diff
@@ -0,0 +1,145 @@
...
+    def delete_helper(id: int) -> Response:
+        return test_client.delete(f"{endpoint}/{id}")
+
+    return delete_helper
+
+
+Post: TypeAlias = Callable[[Any], Response]
+Put: TypeAlias = Callable[[int, Any], Response]
+Get: TypeAlias = Callable[..., Response]
+Delete: TypeAlias = Callable[[int], Response]
```
andrejvelichkovski, line 51: We seem to have the same fixtures defined in `tests/controllers/test_executions/test_reruns.py`. Perhaps we can move them to a shared file and use it in both, maybe in `tests/common.py` or `test/fixtures.py`?

andrejvelichkovski (changes requested):
The code looks good to me, however I left two small notes where I think we can improve the style.

My understanding from your explanation was that we would also have some kind of "confirming issues", so that when someone reports an issue, someone from the team would need to confirm it. Or maybe this was only for reporting issues to environments? I got confused between the two.

backend/test_observer/controllers/test_cases/__init__.py
omar-selo: The test_cases level `APIRouter` would be useful if we wanted to add more endpoints under `test-cases` resource. It's also is more closely similar to how we have setup the `test-executions/reruns` endpoint

backend/tests/controllers/test_cases/test_reported_issues.py
omar-selo: Code wise they are similar. But they represent different pieces of knowledge. These fixtures know how to get, put, delete and post on this particular endpoint. But they don't necessarily work the same way for other endpoints. For reruns for instance, `delete` works differently (it takes a body that contains a list of ids to delete). For test executions (start_test, end_test, etc), put works differently as in it doesn't require an id on path. Admittedly, this shows that TO API is not very uniform. This is something that we should take into consideration when migrating to v2. But as is, it would be difficult to reuse such fixtures across endpoints.

omar-selo:
> My understanding from your explanation was that we would also have some kind of "confirming issues", so that when someone reports an issue, someone from the team would need to confirm it. Or maybe this was only for reporting issues to environments? I got confused between the two.

You are right. As per spec, confirming issues is only for environments. It's assumed that when users create an issue on test cases, it's already confirmed. But environments are treated differently to allow a way for reviewers to tell certlab that something needs checking (e.g. machine is not provisioning, wifi keeps failing, etc.). Once certlab confirmed the issue they'd update it and mark it as confirmed.

backend/tests/controllers/test_cases/test_reported_issues.py
omar-selo: Frankly, while these fixtures do reduce some common code, it's probably arguable whether they're worth it. As in, I wouldn't mind removing them if that's easier to read.

andrejvelichkovski: Oh, I see, I didn't notice the slight differences between the API endpoints we have. I agree about making them more uniform when we migrate to a newer version.

backend/migrations/versions/2024_08_30_1159-6f3eeb521421_add_case_name_to_testcaseissues.py
andrejvelichkovski, line 1: Not really a strong feeling for it, but since all three migrations are really connected to this one feature, what do you think about merging all the migrations in one, to just have one for the whole functionality?

backend/scripts/seed_data.py
```diff
@@ -283,6 +286,24 @@
     ),
 ]
 
+TEST_CASE_ISSUE_REQUESTS = [
+    ReportedIssueRequest(
+        template_id=END_TEST_EXECUTION_REQUESTS[0].test_results[0].name,
```
andrejvelichkovski, line 291: Probably this wont make any difference as it is a seed data script, but I think this should be either `case_name=...` or `template_id=(...).template_id`

andrejvelichkovski (approved):
Thanks for following up on this! Looks really good, excited to see it landed :d

backend/scripts/seed_data.py
```diff
@@ -228,6 +228,7 @@
             ),
             C3TestResult(
                 name="bluetooth4/beacon_eddystone_url_hci0",
+                template_id="luetooth4/beacon_eddystone_url_interface",
```
nadzyah, line 231:
```suggestion
                template_id="bluetooth4/beacon_eddystone_url_interface",
```
I guess it's a typo

backend/test_observer/controllers/test_cases/models.py
```diff
@@ -0,0 +1,26 @@
...
+class ReportedIssueRequest(BaseModel):
+    template_id: str = ""
+    case_name: str = ""
+    description: str
+    url: HttpUrl
+
+    @model_validator(mode="after")
+    def check_a_or_b(self):
+        if not self.case_name and not self.template_id:
+            raise ValueError("either case_name or template_id is required")
```
nadzyah, line 15:
```suggestion
            raise ValueError("Either case_name or template_id is required")
```

nadzyah (approved):

PR #210, opened 2024-09-09
Fetch single artefact on artefact page

omar-selo:
## Description

Makes artefact page fetch single artefact based on id rather than fetching current arteracts. This is beneficial for the following reasons:
- Less data is fetched on artefact page
- Artefact page can show older artefacts even if they aren't on the dashboard (this isn't possible with the old approach)

## Resolved issues

First PR towards resolving https://warthogs.atlassian.net/browse/RTW-342

## Tests

Tested manually locally. Also tested that updating an artefact's status then going back to dashboard page shows the updated status.

### DIFF
--- a/frontend/lib/providers/artefact.dart
+++ b/frontend/lib/providers/artefact.dart
@@ -13,9 +13,17 @@ class Artefact extends _$Artefact {
     return api.getArtefact(artefactId);
   }
 
-  void changeArtefactStatus(id, artefact_model.ArtefactStatus status) async {
+  Future<void> changeArtefactStatus(
+    id,
+    artefact_model.ArtefactStatus status,
+  ) async {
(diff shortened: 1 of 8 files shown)


andrejvelichkovski (changes requested):
Thanks for working on this, but I noticed a small issue while reviewing this, it is that the ratio of completed artefact is not updating as you review environments. It is not a huge issue, but perhaps we can fix it.

Also I think we can clean up a lot of code with this change, in the `providers/family_artefacts` as we have functions to update the artefact status there, and to update the ratio as well, which we probably want to move here.

omar-selo:
Ah good point I completely forgot about the ratio of completed artefact being present in two different locations

andrejvelichkovski (approved):
LGTM!  Thanks for the follow-up

PR #215, opened 2024-09-18
Distinguish environment and test issues to avoid confusion

omar-selo:
## Description

Based on feedback I got from @boukeas, it can be confusing to have both environment and test issues named 'Reported Issues'. Instead naming them different will avoid any mix-up.


boukeas (approved):
Thanks for the change!

PR #217, opened 2024-09-23
Environment issues followup

omar-selo:
## Description

A couple of followup changes from environment issues story:
- Add is_confirmed query parameter to API endpoint
- Make URL only required if the issue is confirmed

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-364

## Tests

Updated backend tests. Then tested front-end manually.


andrejvelichkovski (approved):
This change looks good to me, but I was just thinking that we are sort-of introducing a bit of inconsistency here, as in other data models we treat empty string as no data, such as the Artefact identifiers (series/repo/..) but here we have nullable field as well.

PR #222, opened 2024-10-17
Fix rerun filtered environments doesn't take into consideration review filters

omar-selo:
## Description

Since we have two separate filter groups filtering different lists. We need to only use the inclusive combination.

## Resolved issues

Fixes https://warthogs.atlassian.net/browse/RTW-390

## Tests

Tested manually.

[Screencast from 2024-10-17 15-08-10.webm](https://github.com/user-attachments/assets/00d408c8-0cfb-43cd-9ab3-b6aa21e6afdb)


andrejvelichkovski (approved):
Good catch! Looks good but we would probably need to be more careful if we keep adding more filters here.

andrejvelichkovski (approved):
LGTM

PR #226, opened 2024-11-25
Fix snaps not getting promoted from beta to stable

omar-selo:
## Description

Snaps were not being promoted more than one stage at a time. So when TO saw that a snap has moved to stable even though it was on beta it ignored it.

Note that I've also removed a few entries from the Pydantic models that aren't being used.

## Resolved issues

Fixes https://warthogs.atlassian.net/browse/RTW-369

## Tests

Added a test to confirm that the issue was fixed.

### DIFF
--- a/backend/test_observer/promotion/promoter.py
+++ b/backend/test_observer/promotion/promoter.py
@@ -36,15 +36,6 @@ from test_observer.external_apis.snapcraft import (
 )
 
 logger = logging.getLogger("test-observer-backend")
-
-CHANNEL_PROMOTION_MAP = {
-    # channel -> next-channel
-    "edge": "beta",
-    "beta": "candidate",
-    "candidate": "stable",
-    "stable": "stable",
-}
-
 POCKET_PROMOTION_MAP = {
     # pocket -> next-pocket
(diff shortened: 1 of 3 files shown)


backend/test_observer/promotion/promoter.py
```diff
@@ -150,15 +141,14 @@ def run_snap_promoter(session: Session, artefact: Artefact) -> None:
                 )
                 continue
 
-            next_risk = CHANNEL_PROMOTION_MAP[artefact.stage.name]
             if (
-                risk == next_risk != artefact.stage.name.lower()
+                risk != artefact.stage.name.lower()
```
nadzyah, line 145: is it expected that sometimes an artefact gets promoted skipping one of the stages?

nadzyah (approved):

backend/test_observer/promotion/promoter.py
omar-selo: we've had situations where celery was down or was there some other issue with deployment during which we missed a snap moving from beta to candidate to stable. But I think it's possible to promote a snap skipping a stage through the snapstore

PR #228, opened 2024-11-27
Move rerun button back next to environment name

omar-selo:
## Description

Mixed with the changes to allow multiple reruns on TO, the rerun button was accidentally moved to be next to reruns rather than the environment itself. This PR moves it back.


nadzyah (approved):
Thank you!

PR #231, opened 2024-12-06
Use window function instead of subquery for test execution reports

omar-selo:
## Description

Proposing a slightly different SQL query to retrieve the test executions report. This one uses a window function instead of a subquery that you'd join on. My assumption is that this is faster and more readable.


andrejvelichkovski (approved):
Looks good to me, I agree this will simplify the query

PR #232, opened 2024-12-06
Support test plans

omar-selo:
## Description

Changes:
- Adds test plans to backend and frontend
- Move rerun button from environment to test plan level
- When am execution starts, delete only rerun requests of the same test plan
- Remove rerun filtered environments button as it's not clear which test plans to rerun from each environment
- Split StartTestExecutionRequest model into StartSnapTestExecutionRequest and StartDebTestExecutionRequest. This way the API schema and docs show this distinction
- Remove broken GH workflow of charm integration tests (This workflow never worked)

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-414

## Web service API changes

Start test endpoint now requires an additional key. Namely, `test_plan`. This endpoint is only being used by our jenkins jobs. Those have been updated accordingly already.

## Tests

Added automated tests for backend.

See screenshots for frontend:
Screenshot from 2024-12-06 15-47-37
Screenshot from 2024-12-06 15-48-18


andrejvelichkovski:
Just as we merged the new report in TO, it gives me that idea that maybe we should also expose the `testplan` in the two reporting endpoints?

backend/test_observer/controllers/test_executions/models.py
```diff
@@ -54,6 +54,7 @@ class StartTestExecutionRequest(BaseModel):
     execution_stage: str
     environment: str
     ci_link: Annotated[str, HttpUrl]
+    test_plan: str = Field(max_length=200)
```
andrejvelichkovski, line 57: Do you maybe know if other teams using Test Observer would be providing and using `test_plan`, I wonder if it is a better approach to make this optional and make it default to `""` in case not all teams are providing it, instead of making them explicitly call the API with `"test_plan": ""`. What do you think?

backend/tests/controllers/test_executions/test_start_test.py
```diff
@@ -120,19 +145,9 @@ def test_creates_all_data_models(db_session: Session, execute: Execute):
...
-            "track": "22",
-            "arch": "arm64",
-            "execution_stage": "beta",
-            "environment": "cm3",
-            "ci_link": "http://localhost",
-        },
-    )
+    request = test_request.copy()
+    del request["store"]
+    response = execute(request)
```
andrejvelichkovski, line 150: This test seems to be testing the same behaviour as the newly added `test_required_fields`, do you think we should modify that one to check for `store` as well?

andrejvelichkovski (approved):
Thanks for the changes! I went through all of them and it all seems reasonable to me!

backend/test_observer/controllers/test_executions/models.py
omar-selo: I'd prefer if teams did specify some kind of string here. This is more important in the case of multiple teams running tests on the same environment. We'd need to specify which tests is for which team. We can communicate it to them as this is just a free form string in which they set a name for that kinds of tests they're running

andrejvelichkovski: Fair enough, in that case we can keep the validation in place.

backend/tests/controllers/test_executions/test_start_test.py
omar-selo: Ended up splitting the request models into two. The benefit of that is the API schema and docs now reflect that depending on the family, different requirements are in place

.github/workflows/frontend_charm_integration_tests.yml
andrejvelichkovski, line 1: Thanks for deleting this workflow, we were wasting time for this action that always ended up failing. Should we just file a JIRA ticket to fix this and introduce the fixed integration tests in the repository again?

andrejvelichkovski (approved):
I like the idea that we are separating the start test into two different models based on the family, it reflects how we use TO more realistically. Probably in the future it would be handy if we focus on writing all the `start-tests` parameterized on the two families (snaps/debs).

.github/workflows/frontend_charm_integration_tests.yml
omar-selo: We already have a couple of Jira tickets to add charm integration tests for TO and through CI

PR #233, opened 2024-12-11
Fix test executions stuck in progress

omar-selo:
## Description

Some test executions get stuck "In Progress" if the TF job ended but they didn't receive any results. After some discussion the following approach seems to be the best workaround. Given the following definition of test execution statuses:
- "Passed" a test execution that completed entirely receiving results that are all passing
- "Failed" a test execution that completed entirely receiving results that have some failures
- "Ended Prematurely" a test execution that didn't complete entirely. This could be due to a provision failure, some test hanging, or just a network issue in which TO didn't receive the results.

Then we set a test execution to "Ended Prematurely" when TO receives a job ended event from TF.

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-358

## Tests

Modified the test.


andrejvelichkovski (approved):
LGTM!

PR #234, opened 2024-12-11
Add post test results endpoint

omar-selo:
## Description

Adds a new endpoint to submit test results to a test execution by ID.

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-411

## Web service API changes

New endpoint `POST /v1/test-executions/id/test-results`.
Takes as input a list of test results to add to TO.

## Tests

Added automated tests.

### DIFF
--- a/backend/test_observer/controllers/test_executions/post_results.py
+++ b/backend/test_observer/controllers/test_executions/post_results.py
@@ -1,0 +1,52 @@
+from fastapi import APIRouter, Depends, HTTPException
+from sqlalchemy import delete
+from sqlalchemy.orm import Session
+
+from test_observer.controllers.test_executions.models import TestResultRequest
+from test_observer.data_access.models import TestCase, TestExecution, TestResult
+from test_observer.data_access.repository import get_or_create
+from test_observer.data_access.setup import get_db
+
+router = APIRouter(tags=["test-results"])
+
+
+@router.post("/{id}/test-results")
+def post_results(
+    id: int,
+    request: list[TestResultRequest],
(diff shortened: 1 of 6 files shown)


backend/test_observer/controllers/test_executions/post_results.py
```diff
@@ -0,0 +1,52 @@
...
+            delete(TestResult).where(
+                TestResult.test_execution_id == test_execution.id,
+                TestResult.test_case_id == test_case.id,
+            )
+        )
+
+        test_result = TestResult(
+            test_execution=test_execution,
+            test_case=test_case,
+            status=result.status,
```
rpbritton, line 45: I wonder if the test_execution.status should be recomputed based on the logic in https://github.com/canonical/test_observer/pull/236/files. Or maybe it would make sense that the test_execution.status is just a computed value.

omar-selo: I'm imagining a workflow in which users of TO would do the following:
1. Hit the start test endpoint
2. Submit results through this post endpoint (perhaps multiple times too)
3. End the test execution by sending a patch request with the `COMPLETED` status
In such a workflow if a user hits the endpoint they didn't necessarily finish yet. So perhaps it's best not to change the status of the test execution in case they want to submit more results (at least until they send COMPLETED)

omar-selo: Technically only the `PASSED` and `FAILED` statuses could be computed values. But I don't see a benefit of making them so

PR #237, opened 2024-12-13
Add charm family to test observer

rpbritton:
## Description

Adds charm family and stages to the database, and updates the scattered backend and frontend logic to also support Charms. 

I think it'd be worth making an improvement ticket to centralize all this logic in a file/module just so it's manageable. I don't know when or if there will ever be time to get this done, but having loads of ifs and matches throughout the code is a bit of a headache.

## Resolved issues

[SQT-392](https://warthogs.atlassian.net/browse/SQT-392)

## Documentation

I don't believe any documentation needs to be updated. These changes following the existing pattern used by Snaps and Debs.

## Web service API changes

Not really applicable, can now specify `CHARM` has a family.

## Tests

* Existing unit tests pass
* Charm added to `seed_data.py`
* Test added to the unique artefacts query for Charms that includes all instances of architectures.
* Front end tested by poking around the UI with the seed_data

Screenshot from 2024-12-13 11-46-20
Screenshot from 2024-12-13 11-46-39

[SQT-392]: https://warthogs.atlassian.net/browse/SQT-392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/migrations/versions/2024_12_13_1702-0e09759a33c3_add_charm_family.py
+++ b/backend/migrations/versions/2024_12_13_1702-0e09759a33c3_add_charm_family.py
@@ -9,7 +9,6 @@ from alembic import op
 import sqlalchemy as sa
 
 from test_observer.data_access.models_enums import FamilyName
-from test_observer.data_access.models import Family, Stage
 
 new_families = {
     FamilyName.CHARM: ["edge", "beta", "candidate", "stable"],
@@ -25,32 +24,61 @@ depends_on = None
 
 def upgrade() -> None:
     conn = op.get_bind()
-    session = sa.orm.Session(bind=conn)
+
+    family_table = sa.table(
+        "family",
+        sa.column("id", sa.Integer()),
+        sa.column("name", sa.String()),
+        sa.column("created_at", sa.DateTime()),
+        sa.column("updated_at", sa.DateTime()),
+    )
+    stage_table = sa.table(
+        "stage",
+        sa.column("name", sa.String()),
+        sa.column("family_id", sa.Integer()),
+        sa.column("position", sa.Integer()),
+        sa.column("created_at", sa.DateTime()),
+        sa.column("updated_at", sa.DateTime()),
+    )
 
     for family_name, stage_names in new_families.items():
         # Create family
-        family = Family(name=family_name)
+        (inserted_family,) = conn.execute(
+            sa.insert(family_table)
+            .values(
+                name=family_name,
+                created_at=sa.func.now(),
+                updated_at=sa.func.now(),
+            )
+            .returning(family_table.c.id)
+        )
 
         # Add stages
         stage_position = 10
-        family.stages = []
         for stage_name in stage_names:
-            family.stages.append(Stage(name=stage_name, position=stage_position))
+            conn.execute(
+                sa.insert(stage_table).values(
+                    name=stage_name,
+                    family_id=inserted_family.id,
+                    position=stage_position,
+                    created_at=sa.func.now(),
+                    updated_at=sa.func.now(),
+                )
+            )
             stage_position += 10
-
-        # Add family
-        session.add(family)
-
-    session.commit()
 
 
 def downgrade() -> None:
     conn = op.get_bind()
-    session = sa.orm.Session(bind=conn)
+
+    family_table = sa.table(
+        "family",
+        sa.column("id", sa.Integer()),
+        sa.column("name", sa.String()),
+        sa.column("created_at", sa.DateTime()),
+        sa.column("updated_at", sa.DateTime()),
+    )
 
     # Delete each family, stage deletes on cascade
     for family_name in new_families:
-        family = session.query(Family).filter_by(name=family_name).first()
-        session.delete(family)
-
-    session.commit()
+        conn.execute(sa.delete(family_table).where(family_table.c.name == family_name))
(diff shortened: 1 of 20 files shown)


backend/tests/data_generator.py
```diff
@@ -47,6 +48,7 @@ def gen_user(
     def gen_artefact(
         self,
         stage_name: str,
+        family_name: str = "snap",
```
rpbritton, line 51: This change resulted in a bunch of propagation. Necessary because the testing utilities derived family from the stage, but Charms share stage names with Snaps so that is no longer possible.

rpbritton:
[Ticket to add promoter for Charms](https://warthogs.atlassian.net/browse/SQT-414), I felt it was out of the scope of this PR.

backend/scripts/seed_data.py
```diff
@@ -251,6 +252,18 @@
...
     ),
+    StartCharmTestExecutionRequest(
+        family=FamilyName.CHARM,
+        name="postgresql-k8s",
+        version="123",
+        revision=123,
+        track="14",
+        arch="arm64",
+        execution_stage="candidate",
+        environment="juju=3.5 ubuntu=22.04 cloud=k8s",
```
rpbritton, line 263: As mentioned in the meeting, the environment should be eventually revised to accept multiple values.

backend/test_observer/data_access/repository.py
```diff
@@ -82,39 +82,65 @@ def get_artefacts_by_family(
...
+                    and_(
+                        Artefact.stage_id == subquery.c.stage_id,
+                        Artefact.name == subquery.c.name,
+                        Artefact.created_at == subquery.c.max_created,
+                        Artefact.repo == subquery.c.repo,
+                        Artefact.series == subquery.c.series,
+                    ),
+                )
+
+            case FamilyName.CHARM:
```
rpbritton, line 121: Note the unique query that groups by artefact's build's architecture as well.

backend/test_observer/data_access/models_enums.py
```diff
@@ -25,6 +25,7 @@
 class FamilyName(str, Enum):
     SNAP = "snap"
     DEB = "deb"
+    CHARM = "charm"
```
rpbritton, line 28: No modifications to `models.py`. The `snap_unique` constraint applies to Charms as well (unique on name, version, and track).

However, I would prefer if there was a single unique constraint that applied to all artefacts, would a unique contraint on name, version, and family be adequate?

It's tough to do with the current implementation, but it is possible to add `family_id` as an attribute to `Artefact`, and then a foreign key in `Artefact` on `(Stage.id, Stage.family_id`), and then a unique constraint on `Artefact` for `(name, version, family_id)`.

The caveat there is that I don't believe `Artefact.family_id` is auto populated unless we add an SQL trigger or there's a way that sqlalchemy can, but I didn't look into it that far.

backend/tests/data_access/test_repository.py
```diff
@@ -108,6 +110,75 @@ def test_get_artefacts_by_family_name_latest(
...
+        track="track-1",
+        stage="edge",
+        day=0,
+        arches=["arch-1"],
+    )
+    specs = [
+        # Base case
+        SimpleNamespace(),
+        # Family is not Charm
+        SimpleNamespace(version="2", family="snap", expect=False),
```
rpbritton, line 143: Re: [comment](https://github.com/canonical/test_observer/pull/237/files#r1884292510) to make `Artefact` unique constraint `(name, version, family)`

One thing about the `snap_unique` constraint on `(name, version, track)` is that there cannot be a snap or charm with those fields being the same. Probably never going to arise, but just one reason why it'd be nice to figure out how to include the family on uniqueness.

backend/tests/data_access/test_repository.py
```diff
@@ -108,6 +110,75 @@ def test_get_artefacts_by_family_name_latest(
...
+        if spec.expect:
+            expected_artefacts.append(artefact)
+
+    # Act
+    artefacts = get_artefacts_by_family(db_session, FamilyName.CHARM)
+
+    # Assert
+    assert sorted([a.id for a in artefacts]) == sorted(
+        [a.id for a in expected_artefacts]
+    )
```
omar-selo, line 184: Thanks for adding comprehensive testing. Would it be possible to simplify this test case by splitting it such that each test is checking one particular case at a time?

omar-selo (changes requested):

backend/tests/data_access/test_repository.py
rpbritton: It's an unlikely issue, but could arise if there is a snap and charm with the same name (and version and track).

omar-selo: Good point. We could add a foreign key on Artefact for family_id as you mentioned. Feel free to do it in this PR or some later PR

backend/migrations/versions/2024_12_13_1702-0e09759a33c3_add_charm_family.py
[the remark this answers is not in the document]
rpbritton: Updated to not use ORM.

backend/tests/data_access/test_repository.py
rpbritton: Now split to better identify what is being tested.

omar-selo (approved):
Looks good!

PR #238, opened 2025-01-07
Replace family and stage tables with enum fields

omar-selo:
## Description

Changes:
- Add stage and family fields to artefact entity
- Drop stage and family tables
- Some changes to add user tests to make it faster by avoiding an actual request being made

Reasoning:
Families and stages were first introduced as tables with the idea of it being easier to add families in the future. That turned out not to be so simple though. Adding families requires specific code changes for each family. Hence the code base ended up knowing the families and in fact using them in conditionals. That beats the purpose of storing them as data tables. So this PR removes these tables. Doing so simplified our queries as it avoids the need for joins to figure out an artefact's family or stage.

## Resolved issues

A refactor to help towards https://warthogs.atlassian.net/browse/RTW-301

## Web service API changes

- The name of the first column of reports endpoints has been changed from `Family.name` to `Artefact.family` (these endpoints are only used by us)

## Tests

Updated tests.


backend/test_observer/data_access/models_enums.py
```diff
@@ -23,9 +23,18 @@
...
+    charm = "charm"
+
+
+class StageName(str, Enum):
+    proposed = "proposed"
+    updates = "updates"
+    edge = "edge"
+    beta = "beta"
+    candidate = "candidate"
+    stable = "stable"
```
omar-selo, line 37: We could further distinguish between the stages of each family. But I think that's not necessary right now

backend/test_observer/data_access/models.py
```diff
@@ -108,68 +110,34 @@ def __repr__(self) -> str:
         return data_model_repr(self, "launchpad_handle")
 
 
-class Family(Base):
```
nadzyah, line 111: Am I understanding it correctly that you're replacing these 2 models with just enums?

nadzyah (approved):
This is a big change, but everything looks fine. I've double-checked it :)

backend/pyproject.toml
```diff
@@ -3,6 +3,7 @@ name = "test-observer"
 version = "0.0.0"
 description = "Observe the status and state of certification tests for various artefacts"
 authors = []
+package-mode = false
```
nadzyah, line 6: Perhaps this should be fixed as discussed.

backend/test_observer/data_access/models.py
omar-selo: Yup exactly

backend/pyproject.toml
omar-selo: but TO is not a package so technically it shouldn't be installed. So I opted to set it to false

PR #240, opened 2025-01-09
Fix snap promotion

omar-selo:
## Description

When a snap is available in more than one risk at a time (e.g. edge and beta) TO sets it's stage to the first risk it sees instead of the highest one. This PR fixes the bug.

## Resolved issues

Fixes https://warthogs.atlassian.net/browse/RTW-426

## Tests

Added bug as a test case.

### DIFF
--- a/backend/test_observer/external_apis/snapcraft_models.py
+++ b/backend/test_observer/external_apis/snapcraft_models.py
@@ -19,12 +19,18 @@
 """Mappings for json objects from snapcraft"""
 
 
+from typing import Literal
+
 from pydantic import BaseModel
+
+from test_observer.data_access.models_enums import StageName
 
 
 class Channel(BaseModel):
     architecture: str
-    risk: str
+    risk: Literal[StageName.edge] | Literal[StageName.beta] | Literal[
+        StageName.candidate
+    ] | Literal[StageName.stable]
     track: str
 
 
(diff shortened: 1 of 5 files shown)


backend/test_observer/external_apis/snapcraft_models.py
```diff
@@ -19,12 +19,18 @@
...
 
+from test_observer.data_access.models_enums import StageName
+
 
 class Channel(BaseModel):
     architecture: str
-    risk: str
+    risk: Literal[StageName.edge] | Literal[StageName.beta] | Literal[
+        StageName.candidate
+    ] | Literal[StageName.stable]
```
plars, line 33: Minor/optional comment, I think you can also specify this as something like:
`Literal[StageName.edge, StageName.beta, StageName.candidate, StageName.stable]`
Or even set something like:
```
ValidSnapStages = Literal[StageName.edge, StageName.beta, StageName.candidate, StageName.stable]
...
risk: ValidSnapStages
...
```

plars (approved):
One optional comment below, feel free to ignore though as it's just a subjective thing.
Thanks for the quick fix on this! It's interesting that even before the fix, it seems to eventually make it to the correct channel in TO though.

backend/test_observer/external_apis/snapcraft_models.py
omar-selo: Oh I didn't that `Literal` could take multiple values. I was thinking of using the second approach but I'll delay it to my next PR that should include images as otherwise there will be some merge conflicts

omar-selo:
> it seems to eventually make it to the correct channel in TO though.

Yeah that happens because typically a newer version pops up in the risk it was previously at. So TO would grab the next available channel map that is in a higher risk

PR #241, opened 2025-01-13
Allow null ci link in start-test endpoint

rpbritton:
## Description

Allows the submission to the `/start-test` endpoint of a test execution with a null CI link.

null was already a valid value for a the CI link in the test execution model, but the endpoint would not accept it.

## Resolved issues

[SQT-429][https://warthogs.atlassian.net/browse/SQT-429)

## Documentation

N/A

## Web service API changes

Fully backwards compatible.

## Tests

Unit test added that calls `/start-test` with a null CI link.

[SQT-429]: https://warthogs.atlassian.net/browse/SQT-429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/test_observer/controllers/test_executions/models.py
+++ b/backend/test_observer/controllers/test_executions/models.py
@@ -46,7 +46,7 @@ class _StartTestExecutionRequest(BaseModel):
     arch: str
     execution_stage: StageName
     environment: str
-    ci_link: Annotated[str, HttpUrl]
+    ci_link: Annotated[str, HttpUrl] | None
     test_plan: str = Field(max_length=200)
     initial_status: TestExecutionStatus = TestExecutionStatus.IN_PROGRESS
 
--- a/backend/test_observer/controllers/test_executions/models.py
+++ b/backend/test_observer/controllers/test_executions/models.py
@@ -46,7 +46,7 @@ class _StartTestExecutionRequest(BaseModel):
     arch: str
     execution_stage: StageName
     environment: str
-    ci_link: Annotated[str, HttpUrl] | None
+    ci_link: Annotated[str, HttpUrl] | None = None
     test_plan: str = Field(max_length=200)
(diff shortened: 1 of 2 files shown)


backend/test_observer/controllers/test_executions/models.py
```diff
@@ -46,7 +46,7 @@ class _StartTestExecutionRequest(BaseModel):
     arch: str
     execution_stage: StageName
     environment: str
-    ci_link: Annotated[str, HttpUrl]
+    ci_link: Annotated[str, HttpUrl] | None
```
omar-selo, line 49: Ideally a client wouldn't have to send the ci_link key at all. But to do that, you must set a default value of None. Adding the type hint isn't enough. You can try it out by modifying the test case and dropping the key

backend/tests/controllers/test_executions/test_start_test.py
omar-selo, line 1: I wonder if the [creation of a test execution](https://github.com/canonical/test_observer/blob/main/backend/test_observer/controllers/test_executions/start_test.py#L104-L116) needs modification. It filters by ci_link which can now be None.

omar-selo (changes requested):

backend/test_observer/controllers/test_executions/models.py
rpbritton: Updated to default to None as well as included a test for that.

backend/tests/controllers/test_executions/test_start_test.py
rpbritton: The filter should evaluate to `= NULL` which will always return false, so duplicate calls to `/start-test` without a `ci_link` will create multiple test executions, which I think is reasonable. Added a test case for that scenario as well.

PR #244, opened 2025-01-14
Add image testing backend

omar-selo:
## Description

Adds backend changes required for image artefact family.

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-301

## Web service API changes

Endpoints that return Artefacts now additionally include keys specific to images. This is however not a breaking changes to existing clients.

## Tests

Added some tests and updated others

### DIFF
--- a/backend/test_observer/data_access/models.py
+++ b/backend/test_observer/data_access/models.py
@@ -133,8 +133,7 @@ class Artefact(Base):
     release: Mapped[str] = mapped_column(String(200), default="")
     sha256: Mapped[str] = mapped_column(String(200), default="")
     owner: Mapped[str] = mapped_column(String(200), default="")
-    # TODO add image url
-    # TODO optionally rename prepend family specific fields with family name
+    image_url: Mapped[str] = mapped_column(String(200), default="")
 
     # Relationships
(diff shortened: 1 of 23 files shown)


omar-selo:
worth noting that I tried to create an abstract Artefact class that each family subclasses from (Snap, Deb, Image, Charm). But that ended up creating a huge change and in many places separate logic that could be grouped together. I decided to remove that change

backend/test_observer/data_access/models.py
```diff
@@ -149,7 +154,15 @@ def architectures(self) -> set[str]:
...
-            postgresql_where=column("track") != "",
+            postgresql_where=column("family") == FamilyName.snap.name,
+            unique=True,
+        ),
+        Index(
+            "unique_charm",
+            "name",
+            "version",
+            "track",
+            postgresql_where=column("family") == FamilyName.charm.name,
```
rpbritton, line 165: This is nice, I was trying to do this when adding charms but can't remember why it was causing me grief, something to do with trying to join another table. Was family moved?

rpbritton (approved):
I think all the additions are consistent with the implementations of the other families, and the refactors are great improvements. No questions from me.

PR #245, opened 2025-01-17
Add images to frontend

omar-selo:
## Description

Changes:
- Adds image artefacts page
- Adds family specific artefact filters on each of snaps, debs, charms and images pages
- Use null pattern for artefacts without an assignee and show N/A as assigned user if no user is assigned. This is useful because it allows us to filter artefacts without assignees, make the height of artefact card dynamic without handling the case of a no avatar having a smaller card, and show review progress even without an assignee 

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-302 and https://warthogs.atlassian.net/browse/RTW-336

## Documentation

None

## Web service API changes

None

## Tests

Tested manually as I didn't add any new functionality.

See below video:
[Screencast from 2025-01-17 13-19-23.webm](https://github.com/user-attachments/assets/e0af067c-a6aa-4441-99d1-52fe6072deb1)


frontend/lib/models/filters.dart
```diff
@@ -82,30 +83,55 @@ class Filters<T> with _$Filters<T> {
   }
 }
 
-final emptyArtefactFilters = Filters<Artefact>(
+Filters<Artefact> createEmptyArtefactFilters(FamilyName family) {
```
rpbritton, line 86: +1 refactoring

rpbritton (approved):
Like the other one, straight forward with some good refactoring mixed in, nice! +1

PR #247, opened 2025-01-20
Fix publish frontend workflow by specifying jammy instead of latest ubuntu

omar-selo:
## Description

Charm builds on jammy so the runner building it must run jammy too.

## Resolved issues

Fixes https://warthogs.atlassian.net/browse/RTW-427

### DIFF
--- a/.github/workflows/publish_frontend.yml
+++ b/.github/workflows/publish_frontend.yml
@@ -53,7 +53,7 @@ jobs:
 
   build-and-push-frontend-charm:
     needs: build-and-push-frontend-image
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, linux, xlarge, jammy, x64]
     steps:
       - name: Checkout
         uses: actions/checkout@v2


.github/workflows/publish_frontend.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
   build-and-push-frontend-charm:
     needs: build-and-push-frontend-image
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, linux, xlarge, jammy, x64]
```
andrejvelichkovski, line 56: Do the other options (all apart from jammy) guarantee that they will run jammy?

andrejvelichkovski (approved):
LGTM!

.github/workflows/publish_frontend.yml
omar-selo: Based on the [GH documentation](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job) 

> If you specify an array of strings or variables, your workflow will execute on any runner that matches all of the specified runs-on values. For example, here the job will only run on a self-hosted runner that has the labels linux, x64, and gpu:
> 
> `runs-on: [self-hosted, linux, x64, gpu]`

the existence of `jammy` keyword ensures that the runner will be jammy. It's basically an "and" comparison not "or".

PR #249, opened 2025-01-22
Add sphinx documentation

omar-selo:
## Description

Adds sphinx documentation based on [sphinx-docs-starter-pack](https://github.com/canonical/sphinx-docs-starter-pack). Customizes it and adds appropriate initial documentation.

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-428

## Documentation

Introduces documentation.

Moved root `README.md` into terraform directory and added a simpler one in its place.

## Tests

Adds a few workflows to check docs.


docs/content/explanation/glossary.rst
```diff
@@ -0,0 +1,38 @@
+Glossary
+========
+
+Here is a list of terms used by Test Observer (TO) and what they mean.
+
+Artefact
+--------
+
+An artefact is the thing under test for instance a particular snap or image. At its core an artefact has a name, a version, a family and a stage. Artefacts also have other attributes that are specific to their families (e.g. track is specific to snaps and charms).
```
jocave, line 9:
```suggestion
An artefact is the thing under test, for instance a particular snap or image. An artefact has a name, a version, a family and a stage. Artefacts also have other attributes that are specific to their families (e.g. track is specific to snaps and charms).
```

jocave (changes requested):
One small grammar suggestion, otherwise looks good

docs/index.rst
```diff
@@ -0,0 +1,13 @@
...
+
+Certification currently deploys an instance of TO that they used for reviewing Stable Release Updates (SRUs). Other teams also use this instance for their tests. You can visit `the frontend <https://test-observer.canonical.com/>`_ and view `the API docs <https://test-observer-api.canonical.com/docs>`_, although this currently requires Canonical VPN access. There's also a staging deployment of `frontend <https://test-observer-staging.canonical.com/>`_ and `API <https://test-observer-api-staging.canonical.com/docs>`_ that teams can use to test their integration.
+
+.. toctree::
+   :hidden:
+   :maxdepth: 2
+
+   content/how-to/index
```
tang-mm, line 12: Is there any particular reason for adding the `content/` sub-directory? it will result in an additional URL segment in the built docs e.g. `/content/explanation/...`.  It is not a major issue, just that keeping it consistent with other certification doc sets will make it easier to maintain

tang-mm (approved):
Looks nice. Thanks for setting things up!
We also may introduce some Sphinx directives to make the glossary terms referenceable by other pages, but that can be part of later improvements

docs/index.rst
omar-selo: Good point, I'll remove `content`

jocave (approved):
Thanks for accept the suggestion

PR #250, opened 2025-01-30
Fix and improve the local juju setup

omar-selo:
## Description

Changes:
- On Terraform plan: 
  - fix revisions of dependent charms to the ones currently in production and staging
  - slight bump of juju provider together with switch from series to base application parameters as 0.7 didn't work with the local setup that uses a more recent juju. Terraform however tried to delete the juju model after this bump because now the provider adds cloud information too which was missing. So I had to take the juju model out of the terraform plan. We need to be careful when applying this change as unless we delete the model from terraform state in production then terraform will delete the model with all the applications in it.
  - our postgres deployment doesn't match on staging and production (14/stable [281] vs. 14/candidate [247]). I changed the terraform to 14/stable [281] which is what we have on staging. When tested locally however, terraform could change the channel but not the revision due to a [bug](https://matrix.to/#/!tGlkJIgTBsNijRGgvq:ubuntu.com/$T_Sb7FWDxabN08FUF7cs3U4goH5F-yM8311FVFNATyg?via=ubuntu.com&via=matrix.org). I could however manually upgrade postgres to 281 through juju refresh. So I'll have to do that in production.
- Change the instructions for setting up an environment ready to deploy TO charms through juju and charms. These new instructions and much simpler, more reliable, easier to reproduce, and don't mess up the host machine as they work through multipass VM.
- Added a clout init file to setup the multipass VM

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-433

## Documentation

Updated the documentation

## Tests

Deployed and tested everything locally.


andrejvelichkovski (approved):
I have tested the whole setup and it works as expected

PR #252, opened 2025-01-31
Ensure an up to date OpenAPI schema is included in the repo

mz2:
## Description

Adds a GitHub workflow for checking that there is an up to date copy of the schema under `./backend/schemata`

Rationale: since we run presently copies of the service in Canonical internal networks, for automation purposes it is easier to consume the API this way.

## Resolved issues

None.

## Documentation

README mentions of `pre-commit` added.

## Tests

No production code changes => no tests.


backend/Dockerfile
```diff
@@ -17,4 +17,6 @@ RUN pip3 install poetry && \
 
 COPY . .
 
+HEALTHCHECK --interval=30s --timeout=10s --retries=3 CMD curl --fail http://localhost:30000/v1/version || exit 1
```
omar-selo, line 20: TIL container health checks. Look like a useful addition. One note however, reading the [docs](https://docs.docker.com/reference/dockerfile/#healthcheck) it seems that we should set the exit status to 1 to specify that the health check failed.

backend/README.md
```diff
@@ -10,6 +10,7 @@ This project supports [microk8s](https://microk8s.io/) development environment w
 - Install [microk8s](https://microk8s.io/docs/getting-started) and setup permissions
 - Install [Skaffold](https://skaffold.dev/docs/install/#standalone-binary)
 - Install [Poetry](https://python-poetry.org/docs/#installation)
+- Install [pre-commit](https://pre-commit.com) (best done using `sudo apt install pre-commit`, followed by `pre-commit install` in the root of the repo)
```
omar-selo, line 13: It seems more appropriate to me to have pre-commit yaml and instructions inside `backend` directory. Each directory (backend, frontend) can have its own separate pre-commits and or other kinds of checks. These checks tend to be independent. Even the openapi schema itself would ideally live in backend.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,9 @@
+repos:
+  - repo: local
+    hooks:
+      - id: openapi-fetch-yaml
+        name: Fetch OpenAPI schema
+        entry: backend/scripts/fetch_openapi_schema.sh
+        language: script
+        always_run: true
+        pass_filenames: false
```
omar-selo, line 9: wouldn't it be simpler and more direct to specify which files to watch for here using `files` parameter rather than set `always_run` then use some git trickery to figure out what files changed

backend/scripts/fetch_openapi_schema.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+if git diff --cached --name-only | grep --quiet "backend"; then
+    cd backend
+    poetry install >/dev/null 2>&1
+
+    echo "Starting a transient copy of test-observer-api to fetch the OpenAPI schema..."
+    trap 'kill $(jobs -p)' EXIT
```
omar-selo, line 8: would this trap kill other jobs I could have in the background?

.github/workflows/check_openapi_schema.yml
```diff
@@ -0,0 +1,37 @@
...
+  push:
+    branches: [ main ]
+    paths:
+      - backend/**
+      - ".github/workflows/check_openapi_schema.yaml"
+  pull_request:
+    branches: [ main ]
+    paths:
+      - server/**
+      - ".github/workflows/check_openapi_schema.yaml"
```
omar-selo, line 12: the `pull_request` entry is probably not necessary. Also there is no `server` directory

.github/workflows/check_openapi_schema.yml
omar-selo, line 1: I'd suggest that this workflow lives inside the test_backend workflow as it includes much of the same setup.

omar-selo (changes requested):
Thanks for this addition. It looks good, but I have a few comments. The main one is that it seems more appropriate for the schema and pre-commit to live in backend directory.

backend/README.md
mz2: Makes sense, agreed.

.pre-commit-config.yaml
mz2: Oh, I really just copied what was in hardware-api repo. I'm not very familiar with the tool, can you propose a concrete change?

backend/scripts/fetch_openapi_schema.sh
mz2: No, `jobs -p` would only return the background jobs _from this shell session_. As another line of evidence, I haven't see strangeness when using this as part of the pre-commit flow.

.github/workflows/check_openapi_schema.yml
mz2: Addressed via cd1ea6d8b702b462b0249b5eef18e24e4c0a7eda

.github/workflows/check_openapi_schema.yml
mz2: Addressed via 61a9161c289cfcfe7bccfe188ea764133931bbd1

backend/Dockerfile
mz2: There is `|| exit 1` at the end of the statement there

backend/README.md
mz2: Addressed via a05f09d4e7b02ed6620135515912d6e5d8c1648e

.pre-commit-config.yaml
omar-selo: Something like [this](https://github.com/canonical/test_observer/pull/253)

mz2: Thanks! I've addressed directly on this branch (since I ended up adding a slightly differently drafted README addition on the same lines).

omar-selo (approved):

PR #253, opened 2025-02-03
Use files parameter for pre commit

omar-selo:


mz2:
I took these edits directly into the target branch (slightly different edit to the README with the same meaning)

PR #255, opened 2025-02-07
DTO -> Response renaming

mz2:
## Description

The "DTO" terminology trickles down into the OpenAPI schema generated for the API, and as such it trickles down to also automatically generated clients etc. Feels unnecessary to use an acronym and feels unnecessary to reflect the broad architectural role this entity serves when we can talk about a more specific one (it's specifically a response object).

## Resolved issues

No response

## Documentation

API schema updated.

## Web service API changes

OpenAPI schema adjusted.

## Tests

Doesn't look to have impacted tests (interesting observation itself, but doesn't IMO require further actions in this PR).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .vscode
 
 .idea
+build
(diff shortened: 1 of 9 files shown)


omar-selo (approved):

mz2:
Oops, maybe I hadn't updated the precommit hook on my local working copy? Schema now updated correctly.

PR #268, opened 2025-02-25
Fix postgresql version and correct redis base

omar-selo:
## Description

After deploying the [last Terraform changes](https://github.com/canonical/test_observer/pull/250). I've updated the production postgresql-k8s charm to match the staging one. So it should be safe to fix the revision number in the Terraform plan now. Additionally, turns out redis was deployed using `ubuntu@20.04` base not `ubuntu@22.04` in production and staging. It is not possible to change the base through terraform. So I decided to just set it to what's currently deployed.


andrejvelichkovski (approved):
LGTM!

PR #279, opened 2025-03-11
Fix documentation configuration file and specify extension versions

tang-mm:
## Description

- Fix deprecated variable names in Sphinx configuration file. The deprecation was introduced by `canonical-sphinx` extension v0.4.0, and the warnings will cause RTD doc builds to fail
- Use fixed version number ( currently latest `0.4.0`) on `canonical-sphinx` package to prevent breaking changes

## Resolved issues

NA

## Documentation

NA

## Web service API changes

NA

## Tests
PR build check


omar-selo (approved):
Thanks for the fix!

PR #288, opened 2025-04-02
[SQT-414] Add stage to artefact patch endpoint

rpbritton:
## Description

Adds `stage` as a field that can be modified with a `PATCH /artefacts`

## Resolved issues

[SQT-414](https://warthogs.atlassian.net/browse/SQT-414)

## Web service API changes

* Modifies `PATCH /artefacts` to include `stage` field.

## Tests

Adds test that modifies stage on artefact using `PATCH`.

[SQT-414]: https://warthogs.atlassian.net/browse/SQT-414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/test_observer/data_access/models_enums.py
```diff
@@ -52,6 +52,30 @@ def __ge__(self, other: str) -> bool:
         return self._compare(other) >= 0
 
 
+family_stages: dict[FamilyName, list[StageName]] = {
```
rpbritton, line 55: I think this is good information to centralize, but with how the "start test" logic uses `Literal` I'm not sure how to use it there.

Maybe there is a better way to do this, open to suggestions, but I didn't want to store this information in `controllers/artefacts`.

rpbritton: Maybe this could be a `Dict[FamilyName, Literal[StageName]]`?

rpbritton: I've been playing around with that suggestion for a bit, and maybe I just can't wrap my head around but things seem to get very wonky with the overlapping values.

I can tell having `SnapStage | DebStage | CharmStage | ImageStage` on the `ArtefactPatch` will cause the request type for `edge` to be `SnapStage`, which means custom validation/grouping will needed anyway.

Since there are repeats, I think it makes sense to have the `StageName` enum with all the types of stages (non repeated) and have grouping done after to connect them, like this.

omar-selo: If I understand you correctly, are you proposing we have `StageName` with all the stages in it. Then define `SnapStage`, `DebStage`, `CharmStage`, and `ImageStage` again based on the values of `StageName`? I guess that's fine with me

rpbritton: Added a commit that uses `Literal` for each family stage group, so the stage types can be used statically and dynamically (with `get_args`).

I ran into a problem where SQLAlchemy doesn't support multiple enum types for a column. This was the best way I found to keep all the stage types defined to one enum, and then group them afterwards using `Literal` (as was being done in the start test model, now it's just centralized).

omar-selo: This looks better. But I didn't quite understand why we can't use an enum instead of `Literal` for each stage group. The enum can still be used both for typing and regular usage.

rpbritton: The problem is that neither of these solutions works with SQLAlchemy:

1. Enums are defined like `class SnapStage(str, Enum)`, stage column cannot be `stage: Mapped[SnapStage | DebStage | ...]` (because error about multiple types).
2. Enums are defined like `class SnapStage(StageName)` where `SnapStage` contains `beta`, `edge`, etc. and `StageName` is an empty enum, stage column cannot be `stage: Mapped[StageName]` (because error about `StageName` being empty).

backend/test_observer/controllers/artefacts/artefacts.py
```diff
@@ -136,6 +163,5 @@ def get_artefact_versions(
         .where(Artefact.series == artefact.series)
         .where(Artefact.repo == artefact.repo)
         .where(Artefact.os == artefact.os)
-        .where(Artefact.series == artefact.series)
```
omar-selo, line 139: Good catch, I think this should actually be `.where(Artefact.release == artefact.release)` as it pertains to image testing and is a bug without it.

backend/test_observer/data_access/models_enums.py
```diff
@@ -46,12 +46,36 @@ def __le__(self, other: str) -> bool:
         return self._compare(other) <= 0
 
     def __gt__(self, other: str) -> bool:
-        return self._compare(other) > 1
+        return self._compare(other) > 0
```
omar-selo, line 49: nice catch!

PR #290, opened 2025-04-17
Fix add user action

omar-selo:
## Description

Changes:
- Adds a charm integration test for add_user action
- Fixed the bug where the action wasn't working
- Fixed two other actions that had the same issue (I'm not sure why they used to work but don't anymore)

Note:
I think we can improve on the integrations tests considerably, but there is some good amount of work to get that done. I'd propose we do it in another PR. I actually wasn't able to run the integration tests locally, but they did work through CI.

## Resolved issues

Fixes https://warthogs.atlassian.net/browse/RTW-486

## Tests

Added an integration test for this. As you can see, the test failed properly [here](https://github.com/canonical/test_observer/actions/runs/14519713331/job/40737578198). Then it passed [here](https://github.com/canonical/test_observer/actions/runs/14531508073/job/40772073557) after the fix.


andrejvelichkovski (approved):
LGTM! Thanks for the fix!

PR #291, opened 2025-04-18
Add snap and charm branch

omar-selo:
## Description

Changes:
- Adds branch to Artefact model
- Exposes branch through API endpoints
- Display branch on frontend UI

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-502

## Web service API changes

`GET /v1/artefacts` endpoint now includes `branch`
`GET /v1/reruns` endpoint now includes `branch`
`PUT /v1/test-executions/start-test` endpoint now optionally takes as input `branch`

## Tests

Added appropriate back-end tests. Tested front-end manually, see video below:
[Screencast from 2025-04-18 16-04-49.webm](https://github.com/user-attachments/assets/1d95fa28-78b2-4d5b-9132-c26bd98f2eb8)


backend/test_observer/controllers/test_executions/models.py
```diff
@@ -81,6 +82,7 @@ class StartCharmTestExecutionRequest(_StartTestExecutionRequest):
     family: Literal[FamilyName.charm]
     revision: int
     track: str
+    branch: str | None = ""
```
rpbritton, line 85: If you pass `{"branch": null}` in the request, it would be `None` but then how does that get converted to `""` before being inserted into the database?

rpbritton (approved):
+1 looks good but just a concern about allow `None` as a branch type in the start test request.

backend/test_observer/controllers/test_executions/models.py
omar-selo: Good catch. It is indeed a bug. I would've liked mypy to catch this statically but it didn't because of the use of `getattr`. So refactored the method to avoid using this and indeed mypy flagged it after that.

PR #293, opened 2025-04-22
Add backup restoring db to terraform plan

omar-selo:
## Description

Adds an extra postgres database to the terraform plan and connect it to the s3-integrator. This is make it easy to selectively restore backups on this extra database and extract whatever data we need from it in case of data loss.

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/RTW-431

## Documentation

I plan to add documentation to certification-ops once this is merged.

### DIFF
--- a/terraform/test-observer.tf
+++ b/terraform/test-observer.tf
@@ -199,7 +199,7 @@ resource "juju_integration" "db-backups-restore" {
   model = local.juju_model
 
   application {
-    name = juju_application.pg.name
+    name = juju_application.db-backups-restore.name
   }
 
   application {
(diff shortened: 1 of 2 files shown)


terraform/test-observer.tf
```diff
@@ -182,6 +195,18 @@ resource "juju_integration" "db-backups" {
   }
 }
 
+resource "juju_integration" "db-backups-restore" {
+  model = local.juju_model
+
+  application {
+    name = juju_application.pg.name
```
andrejvelichkovski, line 202: Seems like this juju integration is the same as the one above it, should we use `backup-restoring-db` here instead?

omar-selo: Ah good catch. Typical copy mistake.

andrejvelichkovski (approved):
Thanks for the follow-up!

PR #294, opened 2025-04-22
Fix artefact versions endpoint ignoring branches

omar-selo:
## Description

@boukeas noticed that TO is not considering branches when computing the available versions of an artefact. This PR should fix that.

## Tests

Updated the tests to check for this too.


andrejvelichkovski (approved):
LGTM!

PR #321, opened 2025-06-13
Docker compose based development environment & Dockerfile updates

mz2:
## Description

This PR introduces a Docker Compose based development environment, eventually to replace skaffold, but for now to make getting started with the project as simple as `docker compose up`, and to allow also for CI automation to be developed similar to what is already in place for `canonical/hexr` that checks that the local dev experience does not unexpectedly bit rot.

Associated is also a migration from Poetry to uv package manager (since I was making Dockerfile changes affecting dependency fetching, in order to get seeding working,).

The pre-existing local dev environment is not removed or deprecated with this PR, I figured this would be a downstream step after this workflow has been validated.

## Implementation approach
  - Migrate to uv package manager for consistency with other projects in motion
  - Add Docker Compose setup with automatic database migrations and test data seeding
  - Some tweaks to the charms for readability (I separated out the nginx config to its own resource so that it could be referenced also from the development Dockerfile, but decided not to pool reusing that nginx config yet on this branch, so there's just the charm readability improvement)
- Switch backend and frontend Docker images to use a Canonical provided Ubuntu 24.04 base (upgraded to Noble in the process, since I have been developing both backend and frontend happily on a noble based host directly already for a little while at times)
  
##  Documentation

  - Added CLAUDE.md to guide `claude` based developers in working with the codebase
 - Cover the new local dev workflow in README (no attempt at removing the old dev workflow from docs yet)
  
##  Web service API changes

The `version` endpoint now _can_ return ultimately a [uv-dynamic-versioning](https://pypi.org/project/uv-dynamic-versioning/) based version. This is not done yet in neither production (requires a step in the image build & publish pipeline that sets a wanted version with `UV_DYNAMIC_VERSIONING_BYPASS`) nor docker-compose environment, since the changes on this branch grew sufficiently large that I thought it would be just noise for now. Tbranch sets up the prerequisites for this though, and does not break anything.

## Tests

A GitHub CI pipeline is introduced to run `docker compose up` and check that the service (backend, frontend, db) all turn up healthy.

In order to test locally: `docker compose up` (or `docker-compose up` if a sufficiently old `docker` version is in use locally)

## Tests conducted locally

  - Verified Docker Compose stack starts successfully with docker-compose up
  - Confirmed database migrations run automatically on container startup
  - Tested test data seeding works and can be disabled with SEED_DATA=false

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -1,0 +1,14 @@
+name: Check CLA
+permissions:
+  contents: read
+on: [pull_request]
+
+jobs:
+  cla-check:
+    name: Check CLA
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    steps:
+      - name: Check if CLA is signed
+        uses: canonical/has-signed-canonical-cla@5d1443b94417bd150ad234a82fe21f7340a25e4d # v2


backend/pyproject.toml
```diff
@@ -65,5 +60,8 @@ module = "celery.*"
 ignore_missing_imports = true
 
 [build-system]
-requires = ["poetry-core", "poetry-dynamic-versioning"]
-build-backend = "poetry_dynamic_versioning.backend"
+requires = ["hatchling", "uv-dynamic-versioning"]
```
mz2, line 63: This fixes the versioning visible to the .../version endpoint, which was broken up until now(the backend API server has been reporting to the charm an unchanged 0.0.0).

backend/get_version.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
```
mz2, line 1: This is executed at build time (e.g. at image build time), and can be overridden with `UV_DYNAMIC_VERSIONING_BYPASS` (which is hence added as an `ARG` to the Dockerfile).

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
+version: "3.9"
```
mz2, line 1: A follow-up branch adds to this docker-compose also the containers needed for celery. I thought it was simplest to introduce that (along with integration tests for that) out from this branch that was already quite large.

frontend/web/index.html
```diff
@@ -34,7 +34,7 @@
 
   <script>
     // Actual URL is set at runtime with some nginx config level trickery.
-    window.testObserverAPIBaseURI = 'http://localhost:30000/';
+    window.testObserverAPIBaseURI = 'http://0.0.0.0:30000/';
```
rpbritton, line 37: Why this change? [It seems like 0.0.0.0](https://superuser.com/questions/949428/whats-the-difference-between-127-0-0-1-and-0-0-0-0)) is `is commonly used as a non-routable meta-address used to designate an invalid, unknown or non applicable target (a no particular address placeholder).`, which makes sense to have as an "invalid until set" default, but for the the local dev stuff the default value is used (e.g. in firefox I see backend calls are made to `0.0.0.0:30000`, which admittingly I'm not sure why work).

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+services:
+  test-observer-api:
+    image: test-observer-api
+    build:
+      context: ./backend
+      dockerfile: Dockerfile
+      args:
+        VERSION: ${VERSION:-0.1.2}
+    volumes:
+      - ./test_observer:/test_observer
```
rpbritton, line 11: I think the paths for all these mounts are wrong. In my container they're empty, pretty sure they need a `./backend` prefix on the host side. Why are these being mounted anyway? It doesn't seem like the goal here is to be able to edit code with the containers running (need to docker compose down then up again to rebuild the images).

frontend/web/index.html
mz2: Ah, this is done in the Docker running backend end so that the service binds on all addresses, but is not actually needed in the frontend end, I'll revert this bit.

docker-compose.yml
mz2: Pretty sure it's just claude nonsense you correctly noticed and I didn't -- will fix.

backend/wait_for_app.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
+# Copyright (C) 2023 Canonical Ltd.
```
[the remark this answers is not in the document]
mz2: Because I made the migrations be implicitly run at image build time, so the whole thing comes up with `docker compose up --build` with no other steps. The health check I guess would be sensible separately and would serve a different purpose (it would signal readiness at runtime to the frontend app also?)

backend/pyproject.toml
```diff
@@ -1,42 +1,37 @@
-[tool.poetry]
+[project]
```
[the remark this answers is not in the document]
mz2: a) it's (way) faster, b) it's where new Canonical projects are heading, c) it optionally (and by default) manages the python interpreter version which makes it more foolproof in setting up something like a local dev environment (ofc here we in the Docker images explicitly opt out from that behaviour), and more arguably I guess it follows the relevant PEPs more accurately than poetry does (I can't say I know much about this last bit myself enough to have a point of view on it).

backend/get_version.py
mz2: Yep, I think we should indeed do this. I somewhat postponed addressing this though by not actually making the Docker images posted to the OCI registry in this PR yet.

backend/entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+
+echo "Starting Test Observer Backend..."
+
+# Wait for database to be ready
+echo "Waiting for database to be ready..."
+uv run python wait_for_db.py
+
+# Run database migrations
+echo "Running database migrations..."
+uv run alembic upgrade head
```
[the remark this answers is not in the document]
rpbritton: I think that is a good rename.

backend/entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Get the PID of the uvicorn process
+APP_PID=$!
+
+# Wait for the application to be ready
+echo "Waiting for application to be ready..."
+uv run python wait_for_app.py
+
+# Check if seeding is enabled (default to true for development)
+if [ "${SEED_DATA:-true}" = "true" ]; then
```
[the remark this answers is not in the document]
rpbritton: I suppose, it just feels off to me to have an image that does that by default. I'm not too concerned about it, but just something simple like this to delegate that to docker compose would make more sense to me.
```
services:
  test-observer-backend:
    environment:
      SEED_DATA: true
```

backend/wait_for_app.py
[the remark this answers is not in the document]
mz2: Yep, gotcha. This came out much neater indeed using a healthcheck for the DB and the application.

backend/pyproject.toml
```diff
@@ -57,13 +52,20 @@ lint.select = [
 lint.ignore = ["ANN201", "ANN003", "N999", "ANN204", "N818", "B008"]
 
 [tool.mypy]
-exclude = "charm/*"
```
omar-selo, line 60: why exclude it, is it auto generated?

backend/README.md
```diff
@@ -97,12 +97,12 @@ Two Dockerfiles are provided for the backend application:
 
 ## Versioning
 
-The application is versioned for release purposes using source control tag and commit metadata, using [poetry-dynamic-versioning](https://pypi.org/project/poetry-dynamic-versioning/).
+The application is versioned for release purposes using source control tag and commit metadata, using [setuptools-scm](https://github.com/pypa/setuptools_scm).
```
omar-selo, line 100: I'm confused, I thought it was versioned by https://pypi.org/project/uv-dynamic-versioning/

README.md
```diff
@@ -6,9 +6,41 @@ Certification currently deploys an instance of TO that they used for reviewing S
...
+docker compose down
+
+# Rebuild containers after code changes
+docker compose up --build
+```
+
+This will start:
+
+- **Backend API** at `http://localhost:30000` (with automatic migrations and seeding)
+- **Frontend** at `http://localhost:8080`
```
omar-selo, line 34: I thought the docker compose is setting port to 30001 not 8080

.github/workflows/test_docker_compose.yml
```diff
@@ -0,0 +1,71 @@
...
+      - '.github/workflows/test_docker_compose.yml'
+
+# Cancel in-progress runs if new commit pushed
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test-docker-compose:
+    runs-on: ubuntu-latest
```
omar-selo, line 17: is it ok not to use self hosted runners?

omar-selo (changes requested):
Thanks for the changes, they're all nice quality of life improvements. Although it would be great if future changes were split to much smaller PRs so that the review can be better.

backend/pyproject.toml
```diff
@@ -57,13 +52,20 @@ lint.select = [
...
+exclude = ["charm/*", "set_project_version_hook.py"]
```
omar-selo, line 55: why exclude it, is it auto generated?

frontend/nginx.conf
```diff
@@ -8,6 +8,11 @@ server {
         root   /usr/share/nginx/html;
         index  index.html index.htm;
         try_files $uri $uri/ /index.html =404;
+
+        # Required headers for multi-threaded WASM support
+        # These enable SharedArrayBuffer which is needed for Flutter WASM
+        add_header Cross-Origin-Embedder-Policy require-corp;
+        add_header Cross-Origin-Opener-Policy same-origin;
```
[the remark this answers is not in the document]
mz2: Yep, I removed this in a follow-up commit already.

backend/test_observer/controllers/application/version.py
```diff
@@ -25,6 +27,6 @@
 @router.get("")
 async def get_version():
     try:
-        return {"version": version(__package__)}
+        return {"version": __version__}
```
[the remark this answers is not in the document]
mz2: Ah, that is a good catch!

.github/workflows/test_docker_compose.yml
mz2: It doesn't make much of a difference in this case: both will run it happily, I used this since the project is open source and these runners seem to not have been under crazy demand at least lately.

frontend/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM ubuntu:22.04 as builder
 
 ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update
-RUN apt-get install -y curl git wget unzip libgconf-2-4 gdb libstdc++6 libglu1-mesa fonts-droid-fallback lib32stdc++6 python3
+RUN apt-get install -y curl git wget unzip libgconf-2-4 gdb libstdc++6 libglu1-mesa fonts-droid-fallback python3
```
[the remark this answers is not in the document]
mz2: This was a) OK to do because it was always unnecessary (needed for 32-bit Linux builds, I think?), and b) necessary as part of moving to 24.04 (the package is not there on Noble, probably ABI has changed and it's now sth like `lib32stdc++7`, didn't check?)

docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+    ports:
+      - "30000:30000"
+    depends_on:
+      test-observer-db:
+        condition: service_healthy
+    environment:
+      DB_URL: "postgresql+pg8000://test_observer_user:test_observer_password@test-observer-db:5432/test_observer_db"
+      SEED_DATA: "true"
+      DISABLE_PERIODIC_TASKS: "${DISABLE_PERIODIC_TASKS:-false}"  # Can be overridden for GitHub CI
+      UV_DYNAMIC_VERSIONING_BYPASS: "0.0.0" # this is to be set at image build time for production images
```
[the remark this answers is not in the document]
mz2: The difference is that `VERSION` is nonsense :smile: Removing...

docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+      # Mount source code from host for development
+      - ./backend:/home/app
+      # Preserve virtual environment in anonymous volume
+      - /home/app/.venv
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:30000/v1/version"]
+      interval: 30s
+      timeout: 10s
+      retries: 5
+      start_period: 40s
```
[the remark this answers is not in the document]
mz2: Ah, good point! This is indeed unnecessary.

.github/workflows/test_docker_compose.yml
mz2: I'd propose sticking to these, and moving to the self hosted runners if this particular workflow turns into a bottleneck.

docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+
+  test-observer-frontend:
+    image: test-observer-frontend
+    build:
+      context: ./frontend
+      dockerfile: Dockerfile
+    ports:
+      - "30001:80" # Adjust the port mapping as needed
```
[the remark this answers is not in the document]
mz2: Binding to lower ports would require the process to be root executed

frontend/charm/lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
mz2: We indeed shouldn't be :facepalm: This got somehow included in linter output (I probably run ruff from some other directory than the one that contained the config for it?)

README.md
mz2: You're not wrong :smile:

frontend/README.md
```diff
@@ -41,7 +41,8 @@ $ dart run build_runner build
 Then you can launch the fontend on chrome:
 
 ```bash
-$ flutter run -d chrome
+## You can use --experimental-hot-reload with Flutter versions sufficiently new (3.32.0 or later).
```
mz2, line 44:
```suggestion
# You can use --experimental-hot-reload with Flutter versions >= 3.32.0
```

omar-selo (approved):
Looks good. I left a couple of comments for things that feel a bit annoying. Feel free to ignore them if you don't they're necessary.

PR #325, opened 2025-06-27
ps6 test observer updates

andersson1234:
## Description

This PR includes numerous changes required for using up to date charms in ps6, which was necessary for the deployment of Ubuntu Engineering's test observer instance

## Resolved issues

https://warthogs.atlassian.net/browse/ADT-939

https://warthogs.atlassian.net/browse/ADT-940

https://warthogs.atlassian.net/browse/ADT-942

## Documentation

Documentation changes in a cert ops pr to come

## Web service API changes

No API changes

## Tests

Tested in an environment in prodstack 6.


terraform/test-observer.tf
```diff
@@ -1,18 +1,38 @@
...
+  }
+  info = "Credentials for lego charms"
+}
+
 variable "environment" {
   description = "The environment to deploy to (development, stg, production)"
 }
 
+variable "this_juju_model" {
+  description = "an unfortunate hack."
```
andersson1234, line 28: dropme

andersson1234: or rework me

terraform/test-observer.tf
```diff
@@ -68,18 +94,38 @@ locals {
     stg         = "https://84a48d05b2444e47a7fa176b577bf85a@sentry.is.canonical.com//68",
     development = ""
   }
-  juju_model = "test-observer-${var.environment}"
+  # juju_model = "test-observer-${var.environment}"
```
andersson1234, line 97: drop/rework me

terraform/ps6/test-observer.tf
```diff
@@ -0,0 +1,437 @@
...
+
+
+locals {
+  sentry_dsn_map = {
+    production  = "https://dd931d36e0c24681aaeed6abd312c896@sentry.is.canonical.com//66"
+    stg         = "https://84a48d05b2444e47a7fa176b577bf85a@sentry.is.canonical.com//68",
+    development = ""
+  }
+  httpreq_username = file("${var.config_dir}/httpreq_username")
+  httpreq_password = file("${var.config_dir}/httpreq_password")
```
omar-selo, line 126: does this mean that the `httpreq_password` is stored in plain on the bastion under the `config_dir`. I'm not sure how juju secrets work but if that's the case it might be better to go the vault approach for secrets. On other projects we currently manually add the secret to vault, then the terraform vault plugin just pulls it out for us. Can we do something similar with juju secrets? or perhaps is config file the recommended approach for juju secrets?

terraform/ps6/test-observer.tf
```diff
@@ -0,0 +1,437 @@
...
+  description = "Plugin to use in the lego charm"
+  type = string
+  default = "httpreq"
+}
+
+variable "tls_secret_name" {
+  description = "Secret where the TLS certificate for ingress is stored"
+  type        = string
+  default     = ""
+}
```
omar-selo, line 62: what do you think about splitting "tls_secret_name" into two. One for front-end and one for back-end. Technically, they may not be the same variable.

terraform/ps6/test-observer.tf
```diff
@@ -0,0 +1,437 @@
...
+
+resource "juju_application" "api_ingress_lego" {
+  name = "api-ingress-lego"
+  model = var.juju_model
+  trust = true
+
+  charm {
+    name = "lego"
+    channel = "4/stable"
+  }
```
omar-selo, line 171: can we pin the "revision" number

terraform/ps6/test-observer.tf
```diff
@@ -0,0 +1,437 @@
...
+
+resource "juju_application" "frontend_ingress_lego" {
+  name = "frontend-ingress-lego"
+  model = var.juju_model
+  trust = true
+
+  charm {
+    name = "lego"
+    channel = "4/stable"
+  }
```
omar-selo, line 189: can we pin the revision number

terraform/ps6/test-observer.tf
```diff
@@ -0,0 +1,437 @@
...
+    name    = "test-observer-api"
+    channel = "latest/edge"
+    base    = var.base
+  }
+
+  config = {
+    hostname   = var.environment == "stg" ? "test-observer-api-staging.${var.external_ingress_hostname}" : "test-observer-api.${var.external_ingress_hostname}"
+    port       = var.environment == "development" ? 80 : 443
+    sentry_dsn = local.sentry_dsn_map[var.environment]
+  }
```
omar-selo, line 256: I'd recommend getting rid of `var.environment` and instead adding variables for `hostname`, `port`, and `sentry_dsn`. That seems much more flexible. Especially that I imagine the public version of TO will probably connect to a different sentry_dsn.

terraform/ps6/test-observer.tf
omar-selo, line 1: Feel free to split the terraform file into smaller files as it's growing big

omar-selo (changes requested):
Thanks for all this work. This greatly simplifies our way forward with moving to PS6 too. I tried to be meticulous with the review as it this is a particularly critical piece of configuration.

terraform/ps6/test-observer.tf
```diff
@@ -0,0 +1,437 @@
...
+  description = "Database backups s3-integrator bucket"
+  type        = string
+  default     = ""
+}
+
+variable "restoring_backups_s3_bucket" {
+  description = "Database backups s3-integrator bucket"
+  type        = string
+  default     = ""
+}
```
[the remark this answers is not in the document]
andersson1234: They cannot point to the same bucket, otherwise pg charm fails with something along the lines of:

> cannot backup db, bucket has backups from another cluster

terraform/ps6/test-observer.tf
andersson1234: Good point, I need to set up a sentry_dsn

andersson1234:
I've pushed changes which address most of your comments. I've left a few threads open

andersson1234:
which needs discussing, I'll test this update now

terraform/ps6/test-observer.tf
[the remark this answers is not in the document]
omar-selo: Ah yes this race condition issue. Pointing to different buckets renders `backup-restoring-db` meaningless. So either:
a. we have to manually deploy pg first then import it to the terraform state
b. or we need to figure out another way to extract database backups and then remove `backup-restoring-db` completely

terraform/ps6/test-observer.tf
omar-selo: It seems that it should be easy to add secrets to juju directly through the cli.
```
$ juju help commands | grep secret
add-secret                 Add a new secret.
add-secret-backend         Add a new secret backend to the controller.
grant-secret               Grant access to a secret.
list-secret-backends       Alias for 'secret-backends'.
list-secrets               Alias for 'secrets'.
remove-secret              Remove a existing secret.
remove-secret-backend      Removes a secret backend from the controller.
revoke-secret              Revoke access to a secret.
secret-backends            Lists secret backends available in the controller.
secrets                    Lists secrets available in the model.
show-secret                Shows details for a specific secret.
show-secret-backend        Displays the specified secret backend.
update-secret              Update an existing secret.
update-secret-backend      Update an existing secret backend on the controller.
```

What do you think of adding secrets to juju then referencing them directly without the need of a config file. That way we avoid storing secrets in plain text

terraform/ps6/test-observer.tf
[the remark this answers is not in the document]
andersson1234: I'd like to then keep the `backup-restoring-db` out of this terraform plan then for the time being, is that okay?

terraform/ps6/test-observer.tf
andersson1234: This is also fine with me, will amend!

terraform/ps6/test-observer.tf
andersson1234: (but this is out of scope for this PR)

terraform/ps6/test-observer.tf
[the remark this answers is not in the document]
omar-selo: We need it in cert, but since we're currently using the old terraform plan, this is fine with us. Feel free to remove it

omar-selo (approved):
LGTM! thanks for the contribution!

PR #334, opened 2025-07-09
If needs_assignment is set to false an artefact shouldn't get a due date

omar-selo:
## Description

As pointed out by @boukeas, TO was incorrectly setting a due date for artefacts that need no assignment. Additionally there was a small copy mistake on the front-end.

## Tests

Added automated tests.


boukeas (approved):
Looks good, thanks for fixing this.

PR #337, opened 2025-07-16
Support archiving for snap artefacts

omar-selo:
## Description

Update the snaps promoter to archive snaps that don't have a matching version in the store (because there's a newer version).

## Resolved issues

Resolves the snaps side of https://warthogs.atlassian.net/browse/TO-100. But there's still the debs side.

## Tests

Added and updated tests.


uyarahmet (approved):
Thank you for your contribution! Marking snaps as archived if no matching versions exist in store makes good sense.

PR #342, opened 2025-07-21
Sort test results by execution order

uyarahmet:
## Description

Sort test results by execution order on artifact pages to improve test timeline visibility. Test results now display in chronological execution order rather than arbitrary database order. 

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/SQT-464

## Tests

image

### DIFF
--- a/backend/test_observer/controllers/test_executions/get_test_results.py
+++ b/backend/test_observer/controllers/test_executions/get_test_results.py
@@ -16,6 +16,7 @@
 
 
 from fastapi import APIRouter, Depends, HTTPException
+from sqlalchemy import select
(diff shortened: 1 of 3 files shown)


uyarahmet:
Unfortunately, sorting does bump the time complexity up to O(nlogn) from O(n) but for our typical use case with small result sets the performance difference is negligible.

uyarahmet:
> 2. Can you move the order by to the controller/repository/helper method instead of mode. The test results don't always need to be sorted. It would make queries slower in cases where we don't care about the order

Thank you for your review! Yes, database-level sorting provides cleaner separation of concerns between API and DB as you suggested.

PR #345, opened 2025-07-25
[TS-152] Add execution metadata models and endpoints

rpbritton:
## Description

Adds new execution metadata model and execution metadata management APIs in accordance with [SQ079](https://docs.google.com/document/d/1Kc8sRSoH2TIKTmwZ5AkUYtkGurolaKDAbEl9SOUUBEk/edit).

## Resolved issues

[TO-152](https://warthogs.atlassian.net/browse/TO-152)

## Web service API changes

The OpenAPI schema has been updated, but to summarize:

New endpoints:
- `POST /v1/test-executions/{id}/execution-metadata`
- `GET /v1/test-executions/{id}/execution-metadata`
- `GET /v1/execution-metadata`

There are database migrations as well for the new models.

## Tests

Tested tickets with the unit tests, as well as just hitting the endpoints manually.

[TO-152]: https://warthogs.atlassian.net/browse/TO-152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/test_observer/controllers/execution_metadata/execution_metadata.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+from . import execution_metadata
+
+router = APIRouter(tags=["execution-metadata"])
+router.include_router(execution_metadata.router)
+
+
+@router.get("", response_model=ExecutionMetadataGetResponse)
+def get_execution_metadata(db: Session = Depends(get_db)):
```
rpbritton, line 35: Just as a reminder this endpoint is to support searching for test results across artefacts (and filtering by execution metadata).

backend/test_observer/controllers/execution_metadata/models.py
```diff
@@ -0,0 +1,47 @@
...
+
+from collections import defaultdict
+from collections.abc import Sequence
+
+from test_observer.data_access.models import TestExecutionMetadata
+
+
+class ExecutionMetadata(RootModel[dict[str, list[str]]]):
+    @classmethod
+    def from_rows(cls, values: Sequence[TestExecutionMetadata]) -> "ExecutionMetadata":
```
rpbritton, line 28: Tried a model_validator for auto conversion of this but mypy was complaining about typing and I couldn't figure that out.

backend/test_observer/data_access/models.py
```diff
@@ -322,6 +324,37 @@ class TestExecutionRerunRequest(Base):
...
+
+    category: Mapped[str] = mapped_column(String(200))
+    value: Mapped[str] = mapped_column(String(200))
+
+    test_executions: Mapped[list["TestExecution"]] = relationship(
+        secondary=test_execution_metadata_association_table,
+        back_populates="execution_metadata",
+    )
+
+    __table_args__ = (UniqueConstraint("category", "value"),)
```
omar-selo, line 355: Is it possible to have cascade deletions on TestExecutionMetadata, such that, if a Test Execution is deleted, all related metadata are deleted too if that's the only test execution they're related to.

backend/test_observer/controllers/test_executions/execution_metadata.py
```diff
@@ -0,0 +1,130 @@
...
+            test_execution.execution_metadata
+        ),
+    )
+
+
+@router.post(
+    "/{test_execution_id}/execution-metadata",
+    response_model=ExecutionMetadataGetResponse,
+)
+def post_execution_metadata(
```
omar-selo, line 58: Given that `ExecutionMetadataGetResponse` can be returned from this post endpoint, perhaps it might be better named as just `ExecutionMetadataResponse`

backend/test_observer/controllers/test_executions/execution_metadata.py
```diff
@@ -0,0 +1,130 @@
...
+    # Unpack metadata into list
+    execution_metadata_rows = request.execution_metadata.to_rows()
+
+    # Exit if none given
+    if len(execution_metadata_rows) == 0:
+        return ExecutionMetadataGetResponse(
+            execution_metadata=ExecutionMetadata.from_rows(
+                test_execution.execution_metadata
+            ),
+        )
```
omar-selo, line 77: It seems unexpected for me that a post returns objects it didn't create. It's like posting to `/v1/users` then getting as a response all the users

backend/test_observer/controllers/test_executions/execution_metadata.py
```diff
@@ -0,0 +1,130 @@
...
+)
+def get_execution_metadata(test_execution_id: int, db: Session = Depends(get_db)):
+    test_execution = db.get(TestExecution, test_execution_id)
+    if test_execution is None:
+        raise HTTPException(status_code=404, detail="TestExecution not found")
+    return ExecutionMetadataGetResponse(
+        execution_metadata=ExecutionMetadata.from_rows(
+            test_execution.execution_metadata
+        ),
+    )
```
omar-selo, line 51: We could also just return metadata with the TestExecution object elsewhere in the code. That would be better if the front-end needs to show metadata for multiple test executions at the same time under an artefact. Otherwise this is fine too. Not sure we will use this endpoint though.

backend/test_observer/controllers/execution_metadata/models.py
omar-selo, line 1: Since "ExecutionMetadataGetResponse" doesn't include test executions, how would you query the API for test executions that have a certain set of metadata? Or is this to be added later on in the future?

omar-selo (changes requested):

backend/test_observer/data_access/models.py
rpbritton: I don't believe so, at least not without some sort of manual check, since it's a many to many mapping (e.g. since one test execution metadata can be mapped to multiple test executions, when one of those test executions is removed there isn't a way to check the metadata isn't mapped to anything else).

backend/test_observer/controllers/test_executions/execution_metadata.py
rpbritton: I think this endpoint would be called when you click to expand a test execution, but I suppose I could add it to the response in `GET /v1/artefacts/{id}/builds`, and that way we'd have it for filtering too. Good idea, I'll implement that

backend/test_observer/controllers/test_executions/execution_metadata.py
rpbritton: Updated to return the `TestExecutionResponse`, which may clearer? It's still returning objects it didn't create (e.g. the whole Test Execution), but I think it's a reasonable design decision. Like it shows the side effects for the POST.

backend/test_observer/controllers/execution_metadata/models.py
omar-selo: I see, so the GET test results endpoint will allow for metadata filters too.

backend/test_observer/controllers/test_executions/execution_metadata.py
omar-selo: This also feels unexpected for me. Specifically the parent object being returned when I post to child. This seems much closer to a PATCH on `/v1/test-executions/id/` where you pass `execution_metadata` field. It works well also symmetrically. In the sense you're dealing with TestExecution and this just happens to be a child property whether it was fetched or added. wdyt of the PATCH approach?

omar-selo (approved):
Looks good!

PR #349, opened 2025-08-06
[Fix] Check length when posting execution metadata

rpbritton:
## Description

This PR ensures the lengths of execution metadata categories and values are between 1 and 200 characters.

## Resolved issues

Attempting to create execution metadata longer than 200 characters causes Test Observer to return a `500` error (as it fails to insert into database).

## Documentation

N/A

## Web service API changes

I don't think this can show up in the swagger? At least with how execution metadata is a root model?

## Tests

Tests added verifying categories and values and invalid lengths return 422.


omar-selo (changes requested):
It dawned on me as I was reviewing Ahemt's last PR that the `RootModel` approach has many drawbacks. Namely, openapi schema will be very lacking. So the API docs wont show anything useful about the schema, and an auto generated client of the API will be equally clueless. I'd prefer we take a different approach for `ExecutionMetadata` even if it involves changing the schema.

rpbritton:
@omar-selo I figured it out using `Annotated`. Now the openapi schema has actually good documentation for this type (as well as using `max_length` and `min_length` to let pydantic do the validation).

omar-selo (approved):
Nice, I didn't know we can do that, glad it works

PR #351, opened 2025-08-08
[TO-147] Add attachment rule models

rpbritton:
## Description

Adds models for attachment rules, rules that automatically attach issues to test results based on attributes of the test result, and the API endpoints to create, modify, and delete them. This PR includes the follow attributes for matching:

- Family
- Environment name
- Test case name
- Template id
- Execution Metdata

## Resolved issues

[TO-147](https://warthogs.atlassian.net/browse/TO-147)

## Documentation

OpenAPI reflects endpoint changes.

## Web service API changes

- `POST /issues/{id}/attachment-rules`: Create new attachment rules. Not indepotent. May create multiple attachment rules with the same set of criteria.
- `PATCH /issues/{id}/attachment-rules/{id}`: Modify an attachment rule (only enable or disable allowed).
- `DELETE /issues/{id}/attachment-rules/{id}`: Delete the attachment rule.

## Tests

Unit tests added for all the endpoints which verify the models too.

[TO-147]: https://warthogs.atlassian.net/browse/TO-147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/test_observer/controllers/issues/attachment_rules.py
+++ b/backend/test_observer/controllers/issues/attachment_rules.py
@@ -45,7 +45,7 @@ def post_attachment_rule(
     # Get the issue
     issue = db.get(Issue, issue_id)
     if issue is None:
-        raise HTTPException(status_code=404, detail="Attachment rule not found")
+        raise HTTPException(status_code=404, detail="Issue not found")
 
     # Create the attachment rule
     attachment_rule = IssueTestResultAttachmentRule(
--- a/backend/test_observer/controllers/issues/attachment_rules.py
+++ b/backend/test_observer/controllers/issues/attachment_rules.py
@@ -87,7 +87,7 @@ def patch_attachment_rule(
         raise HTTPException(status_code=404, detail="Attachment rule not found")
     if attachment_rule.issue_id != issue_id:
         raise HTTPException(
-            status_code=403, detail="Attachment rule not attached to given issue"
+            status_code=400, detail="Attachment rule not attached to given issue"
         )
 
     # Modify the attachment rule
@@ -112,7 +112,7 @@ def delete_attachment_rule(
         return
     if attachment_rule.issue_id != issue_id:
         raise HTTPException(
-            status_code=403, detail="Attachment rule not attached to given issue"
+            status_code=400, detail="Attachment rule not attached to given issue"
         )
 
     # Delete the attachment rule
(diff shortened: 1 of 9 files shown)


rpbritton:
I want to provide some context to a design consideration, but I'm not dead set on my direction for this.

The question is whether to reuse existing tables for attachment rules or define separate attributes.

1. The existing TestExecutionMetadata table records factual information about test runs. It is owned by those runs and deleted if the run is removed. Rule metadata differs as it defines conditions or filters that may match current or future test executions. Combining these purposes would blur the semantic meaning and complicate the data model.

2. Because TestExecutionMetadata is tied to test runs, deleting runs could unintentionally remove metadata required for rules if the same table is shared. This creates potential lifecycle management issues.

3. This concern extends beyond metadata to all matching attributes such as test case name, template id, and environment name. For example, environment names cannot reliably use foreign keys because uniqueness depends on the combination of architecture and name, making foreign keys impractical for some attributes.

4. Rules may need to specify conditions for attributes that have not yet appeared in any test executions. Maintaining separate rule attributes supports clearer semantics and safer data management.

This differs from the original spec of attachment rules, where foreign keys would be used.

backend/test_observer/data_access/models.py
omar-selo, line 1: Note, please add indexes to foreign keys for faster joins.

backend/test_observer/controllers/issues/attachment_rules.py
```diff
@@ -0,0 +1,120 @@
...
+    db: Session = Depends(get_db),
+):
+    # Get the attachment rule
+    attachment_rule = db.get(IssueTestResultAttachmentRule, attachment_rule_id)
+    if attachment_rule is None:
+        raise HTTPException(status_code=404, detail="Attachment rule not found")
+    if attachment_rule.issue_id != issue_id:
+        raise HTTPException(
+            status_code=403, detail="Attachment rule not attached to given issue"
+        )
```
omar-selo, line 91: 403 is not authorized. Technically the user may be authorized to disable or enable the attachment rule, but has used the incorrect issue id. This is why this seems more like a 400 for me. The user just made an invalid request.

omar-selo (changes requested):
I agree with your design decision, we should definitely not reference rule data with foreign keys for attachment rules. I just left a few comments on other parts.

PR #353, opened 2025-08-13
[TO-149] Add attachment rule logic after test result post

rpbritton:
## Description

When test results are posted, attach any issues specified by any matching and enabled attachment rules.

I considered trying to hook this in to run in more places (even looking at the `after_update` ORM events), but I believe that could give way to unexpected edge cases. Such as if you match on environment and the environment name was somehow changed, are attachment rules rerun? Also possible performance implications if run too much/in unexpected places.

I think asserting that attachment rules run once at the time of posting is a reasonable and reliable way for this to function. Note that it is important to ensure that everything that needs to be updated is updated, e.g. for Charm QA work, we need to now post the execution metadata *before* posting test results.

## Resolved issues

[TO-149](https://warthogs.atlassian.net/browse/TO-149)

## Web service API changes

OpenAPI schema is updated with just a semantic change. APIs did not change (with the exception that now attachment rules are run after results are posted).

## Tests

Tests have been added for all new features.

[TO-149]: https://warthogs.atlassian.net/browse/TO-149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/test_observer/controllers/issues/attachment_rules_logic.py
+++ b/backend/test_observer/controllers/issues/attachment_rules_logic.py
@@ -23,8 +23,6 @@ from sqlalchemy import (
     func,
     exists,
     and_,
-    type_coerce,
-    Enum,
     ColumnElement,
     Select,
     literal,
@@ -39,15 +37,14 @@ from test_observer.data_access.models import (
     TestExecution,
     IssueTestResultAttachment,
 )
-from test_observer.data_access.models_enums import FamilyName
 
 
 def _array_empty_or_contains(
-    array: InstrumentedAttribute, value: object
+    array: InstrumentedAttribute, value: ColumnElement
 ) -> ColumnElement[bool]:
     return or_(
         func.cardinality(array) == 0,
-        array.any(value),  # type: ignore[arg-type]
+        array.any(value),
     )
 
 
@@ -123,10 +120,7 @@ def query_matching_test_result_attachment_rules(
     stmt = stmt.where(
         _array_empty_or_contains(
             IssueTestResultAttachmentRule.families,
-            type_coerce(
-                test_result.test_execution.artefact_build.artefact.family,
-                Enum(FamilyName),
-            ),
+            literal(test_result.test_execution.artefact_build.artefact.family),
         )
     )
 
@@ -134,14 +128,15 @@ def query_matching_test_result_attachment_rules(
     stmt = stmt.where(
         _array_empty_or_contains(
             IssueTestResultAttachmentRule.environment_names,
-            test_result.test_execution.environment.name,
+            literal(test_result.test_execution.environment.name),
         )
     )
 
     # Filter test_case_names
     stmt = stmt.where(
         _array_empty_or_contains(
-            IssueTestResultAttachmentRule.test_case_names, test_result.test_case.name
+            IssueTestResultAttachmentRule.test_case_names,
+            literal(test_result.test_case.name),
         )
     )
 
@@ -149,7 +144,7 @@ def query_matching_test_result_attachment_rules(
     stmt = stmt.where(
         _array_empty_or_contains(
             IssueTestResultAttachmentRule.template_ids,
-            test_result.test_case.template_id,
+            literal(test_result.test_case.template_id),
         )
     )
 


backend/test_observer/controllers/issues/models.py
```diff
@@ -72,36 +70,6 @@ class IssueTestResultAttachmentRulePatchRequest(BaseModel):
     enabled: bool | None = None
 
 
-class IssueTestResultAttachmentRuleResponse(BaseModel):
```
rpbritton, line 75: Moved to `shared_models.py` so it can be included in `MinimalIssueTestResultAttachmentResponse`

backend/test_observer/data_access/models.py
```diff
@@ -691,12 +691,11 @@ class IssueTestResultAttachmentRuleExecutionMetadata(Base):
     A table for attachment rules to match on execution metadata
     """
 
-    __test__ = False
     __tablename__ = "issue_test_result_attachment_rule_execution_metadata"
+    __table_args__ = (UniqueConstraint("attachment_rule_id", "category", "value"),)
 
     attachment_rule_id: Mapped[int] = mapped_column(
-        ForeignKey("issue_test_result_attachment_rule.id", ondelete="CASCADE"),
-        primary_key=True,
```
rpbritton, line 699: Bad primary key here, as it is a composite primary key with `id` and a bit meaningless.

backend/test_observer/controllers/issues/attachment_rules_logic.py
```diff
@@ -0,0 +1,162 @@
...
+    # Filter template_ids
+    stmt = stmt.where(
+        _empty_or_contains(
+            IssueTestResultAttachmentRule.template_ids,
+            test_result.test_case.template_id,
+        )
+    )
+
+    # Filter execution metadata
+    unmatched_categories = (
```
rpbritton, line 98: Rough idea here is find all the attachment rule categories that are unfulfilled by the test execution's execution metadata. (For all categories in the attachment rule, at least one of the category/value pairs must exist in the test execution's metadata).

Then only select attachment rules that have no unmatched categories. Should be reasonably performant.

omar-selo: I tried but I'm afraid I couldn't get this part of the code. I can try to squeeze my brain harder to get it, but I'd prefer if it can be simplified to ease readability. We know that we're only interested in the metadata of one test execution so we can just get them through simply `test_result.test_execution.execution_metadata`. We can also build on the original stmt above by adding a join to `IssueTestResultAttachmentRuleExecutionMetadata`. Can't we then just check that all the categories in `test_result.test_execution.execution_metadata` are in `TestExecutionMetadata.execution_metadata` with at least one value existing? This seems more straightforward to me. But I haven't worked out how to convert it to SQL so it could be worse.

backend/test_observer/controllers/issues/attachment_rules_logic.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def apply_test_result_attachment_rules(db: Session, test_result: TestResult):
+    # Fetch matching attachment rule
+    attachment_rules_stmt = query_matching_test_result_attachment_rules(test_result)
+
+    # Apply attachment rules based on creation
+    attachment_rules_stmt = attachment_rules_stmt.order_by(
+        IssueTestResultAttachmentRule.created_at
+    )
```
omar-selo, line 143: for ordering you can use `id` instead of `created_at` as the former has an index while the latter doesn't

backend/test_observer/controllers/issues/attachment_rules.py
```diff
@@ -26,16 +26,18 @@
...
 router = APIRouter()
 
 
 @router.post(
-    "/{issue_id}/attachment-rules", response_model=IssueTestResultAttachmentRuleResponse
+    "/{issue_id}/attachment-rules",
+    response_model=MinimalIssueTestResultAttachmentRuleResponse,
 )
 def post_attachment_rule(
     issue_id: int,
```
omar-selo, line 43: Q. When a user creates a new attachment rule, should existing test results passing that rule get attached automatically. Or are attachment rules only for future test results?

omar-selo (changes requested):

backend/test_observer/controllers/issues/attachment_rules_logic.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def _empty_or_contains(
+    array: InstrumentedAttribute, value: object
+) -> ColumnElement[bool]:
+    return or_(
+        func.cardinality(array) == 0,
+        value == any_(array),
+    )
+
```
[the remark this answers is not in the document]
rpbritton: Updated to use a more conventional `array.any(value)`. That actually fixes the type_coerce thing too.

backend/test_observer/controllers/issues/attachment_rules.py
rpbritton: When the user creates an attachment rule, there will be a checkbox for whether or not to check historical runs for matches.

backend/test_observer/controllers/issues/attachment_rules_logic.py
```diff
@@ -0,0 +1,157 @@
...
+    stmt = select(IssueTestResultAttachmentRule)
+
+    # Only fetch enabled rules
+    stmt = stmt.where(IssueTestResultAttachmentRule.enabled)
+
+    # Filter families
+    stmt = stmt.where(
+        _empty_or_contains(
+            IssueTestResultAttachmentRule.families,
+            type_coerce(
```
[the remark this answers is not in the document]
rpbritton: Was able to remove it by using `array.any(value)`, which correctly determined the type.

PR #359, opened 2025-08-26
Fix incorrect handling of hostname variables & local SAML setup

omar-selo:
## Description

There were a couple of issues with how the passed hostname terraform variables was being used. This PR fixes them to help deploy the SAML authentication changes properly.

Additionally, I noticed that the last commit broke the local SAML authentication setup. So I included a fix with this.


uyarahmet (approved):
LGTM!

PR #364, opened 2025-09-02
Support frontend auth

omar-selo:
## Description

Adds server side user sessions through signed session cookies together with some CSRF protection in the form of CORS and X-CSRF-TOKEN header. Additionally:
- Now a log-in creates a user in TO database if not already there
- Automatically fetches launchpad handle on user creation
- Periodic celery task that deletes expired user sessions
- Front-end "Log in" button and changes to send session cookies
- Add "is_reviewer" key to users as not all users should be assigned to artefacts

Note that there is still more work to do for authentication namely, the ability to log out and application authorization. This PR is just a step forward enabling web auth sessions.

## Resolved issues

Towards resolving https://warthogs.atlassian.net/browse/TO-170

## Web service API changes

No breaking changes, but a user object response now exposes `email` and `launchpad_email` is deprecated. Also log-in endpoint now returns a session cookie. There's also a new `/v1/users/me`.

## Tests

Updated SAML integration tests and added some unit tests.

See video below for full front-end flow.
[Screencast from 2025-09-01 17-29-52.webm](https://github.com/user-attachments/assets/6bbc41f8-fca3-4d92-8bf7-9fbc74d2d68e)

### DIFF
--- a/frontend/.vscode/launch.json
+++ b/frontend/.vscode/launch.json
@@ -1,0 +1,28 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "frontend",
+            "request": "launch",
+            "type": "dart",
+            "args": ["--web-port", "30001"]
+        },
+        {
+            "name": "frontend (profile mode)",
+            "request": "launch",
+            "type": "dart",
+            "flutterMode": "profile",
+            "args": ["--web-port", "30001"]
+        },
+        {
+            "name": "frontend (release mode)",
+            "request": "launch",
+            "type": "dart",
+            "flutterMode": "release",
+            "args": ["--web-port", "30001"]
+        }
+    ]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
-# Visual Studio Code configurations
-.vscode
-
 .idea
 build
 


backend/saml/simplesamlphp/authsources.php
```diff
@@ -15,6 +15,11 @@
             'lp_teams' => array('canonical-hw-cert'),
             'email' => 'certbot@canonical.com'
         ),
+        'omar-selo:password' => array(
+            'fullname' => 'Omar Abou Selo',
+            'lp_teams' => array('canonical'),
+            'email' => 'omar.selo@canonical.com'
+        ),
```
omar-selo, line 22: Added this user for the integration test that actually fetches from launchpad API

backend/test_observer/controllers/artefacts/models.py
```diff
@@ -43,8 +43,12 @@ class UserResponse(BaseModel):
...
 
     id: int
-    launchpad_handle: str
-    launchpad_email: str
+    launchpad_handle: str | None = None
+    email: str
+    launchpad_email: str = Field(
+        validation_alias=AliasPath("email"),
+        deprecated="launchpad_email is deprecated please use email instead",
+    )
```
omar-selo, line 51: kept `launchpad_email` for backwards compatibility as it might be used by the kernel team

backend/test_observer/controllers/auth/saml.py
```diff
@@ -148,6 +190,11 @@ async def _prepare_from_fastapi_request(request: Request) -> dict[str, Any]:
         "get_data": {},
     }
 
+    logger.info("Received a login request with the following parameters:")
+    logger.info(f"request.headers {request.headers.items()}")
+    logger.info(f"request.url.scheme {request.url.scheme}")
+    logger.info(f"result['server_port'] {result['server_port']}")
```
omar-selo, line 196: Currently the authentication workflow is not working on staging and production, it falls apart when TO gets the response from U1. SAML validation requires the response to be sent to `https://test-observer-api.canonical.com/v1/auth/saml/acs` but it gets `http://test-observer-api.canonical.com/v1/auth/saml/acs` instead (notice "https" -> "http"). I can't replicate the issue locally sadly. But I suspect it's from this function. So I'm adding these logs temporarily.

backend/test_observer/data_access/models.py
```diff
@@ -92,14 +92,31 @@ class User(Base):
     # user is a reserved name in PostgreSQL
     __tablename__ = "app_user"
 
-    launchpad_email: Mapped[str] = mapped_column(unique=True)
-    launchpad_handle: Mapped[str]
+    email: Mapped[str] = mapped_column(unique=True)
+    launchpad_handle: Mapped[str | None] = mapped_column(default=None)
     name: Mapped[str]
+    is_reviewer: Mapped[bool] = mapped_column(default=False)
```
omar-selo, line 98: After deploying, I'll need to set all current users to reviewers manually through the db. I don't think this should be a db migration since it's specific to our deployment. But it could be since the public deployment doesn't have users yet.

backend/test_observer/controllers/auth/saml.py
```diff
@@ -107,12 +117,44 @@ async def saml_login_callback(request: Request):
...
+        if frontend_url.netloc == return_url.netloc:
+            response = RedirectResponse(
+                url=return_to, status_code=status.HTTP_302_FOUND
+            )
+            return response
+        else:
+            # Note by default python3-saml returns to API url
+            logger.warning(f"Received invalid return url {return_to}")
+
+    return user
```
nadzyah, line 142: I'm a bit concerned about the return type inconsistency. If the URL for redirect exists and is valid, we return response, otherwise, we return a user object. What would the user get if the URL is incorrect in this case? and what if the response if fine, how do we get the user information?

backend/saml/simplesamlphp/authsources.php
nadzyah: the user could be `mark` to keep the same vibe as LP has as its suggestions for default username

backend/tests/users/test_change_assignee.py
```diff
@@ -25,12 +25,12 @@
 def test_change_assignee(db_session: Session, generator: DataGenerator):
     artefact = generator.gen_artefact(StageName.beta)
     user1 = generator.gen_user(
-        launchpad_email="user1@email.com",
+        email="user1@email.com",
```
nadzyah, line 28: I would recommend using the `example.com` domain name instead

backend/test_observer/controllers/auth/saml.py
omar-selo: Good point. The user returned is actually just for testing, there's no need for it. But it must be the case that we either redirect or not as we may not have a redirect url or it might be a redirect attack. I've removed the user return statement

frontend/.vscode/launch.json
omar-selo, line 1: @nadzyah I pushed this vscode launch file to ensure that whenever someone launches flutter from vscode the port is set to 30001. This is because CORS on back-end is now more strict and the port has to match

nadzyah (approved):

PR #371, opened 2025-09-09
Add team permissions

omar-selo:
## Description

This PR adds permissions and useful endpoints for viewing and managing permissions and teams.

Changes:
- Add permissions to teams
- Add endpoints for teams
- Add an endpoint for permissions
- Add endpoints for users

## Resolved issues

A step towards resolving https://warthogs.atlassian.net/browse/TO-174

## Documentation

I need to write documentation for all of auth once I'm done with it.

## Tests

Added automated tests.

### DIFF
--- a/backend/tests/controllers/permissions/test_permissions.py
+++ b/backend/tests/controllers/permissions/test_permissions.py
@@ -17,8 +17,10 @@
 
 from fastapi.testclient import TestClient
 
+from test_observer.common.permissions import Permission
+
 
 def test_get_permissions(test_client: TestClient):
     response = test_client.get("/v1/permissions")
     assert response.status_code == 200
-    assert response.json() == []
+    assert response.json() == [p.value for p in Permission]


backend/test_observer/controllers/teams/models.py
```diff
@@ -0,0 +1,38 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from pydantic import BaseModel
+
+from test_observer.common.permissions import Permission
+
+
+class UserMinimalResponse(BaseModel):
+    id: int
+    launchpad_handle: str | None = None
```
nadzyah, line 24: can it be that a user doesn't have a LP ID?

backend/test_observer/controllers/teams/models.py
```diff
@@ -0,0 +1,38 @@
...
+
+class TeamResponse(BaseModel):
+    id: int
+    name: str
+    permissions: list[str]
+    members: list[UserMinimalResponse]
+
+
+class TeamPatch(BaseModel):
+    permissions: list[Permission] | None = None
```
nadzyah, line 38: I'm curious why anybody would submit a patch request with no data (the only field permissions set to None)

backend/test_observer/controllers/teams/models.py
omar-selo: Yes, because not every ubuntu one account has a corresponding LP account

backend/test_observer/controllers/teams/models.py
omar-selo: This is more for future proofing. As in if we add more properties to update to the endpoint, we wont have to change this to be optional

PR #375, opened 2025-09-10
[TO-158] Add execution metadata to test result search api

rpbritton:

### DIFF
--- a/backend/test_observer/controllers/test_results/test_results.py
+++ b/backend/test_observer/controllers/test_results/test_results.py
@@ -15,10 +15,21 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 from datetime import datetime
-from typing import Annotated
-from fastapi import APIRouter, Depends, Query
-from sqlalchemy import and_, desc, func, select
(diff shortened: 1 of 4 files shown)


backend/tests/controllers/test_results/test_test_results.py
```diff
@@ -635,162 +738,6 @@ def test_ordering_consistency(
             assert len(our_results) == 2
 
 
-class TestGetEnvironments:
```
rpbritton, line 638: I don't believe these belong here? And they look to be duplicated by the tests that do exist where I'd expect them to be.

backend/test_observer/controllers/test_results/test_results.py
```diff
@@ -63,6 +163,9 @@ def search_test_results(
     template_ids: Annotated[
         list[str] | None, Query(description="Filter by template IDs (comma-separated)")
```
rpbritton, line 164: Note, I saw that `comma-separated` isn't really true here. Should I remove these comments? FYI the execution metadata does support passing in comma-separated values, but I can remove that for consistency.

PR #389, opened 2025-09-18
[TO-151] Add bulk attach issues from filters

rpbritton:
## Description

Modifies the issue attach and detach endpoints to accept the use of test result filters. This will be a better way to bulk attach or detach issues from the test results page, or when attaching an issue.

## Resolved issues

Previous to use bulk attach issues, one would need to fetch all the test results in the frontend.

## Web service API changes

OpenAPI updated to match.

## Tests

All existing `GET /test-results` endpoint functionality is the same, so all those existing tests pass as expected.
Added tests for `GET /issues/{id}/attach` and `GET /issues/{id}/detach` when using test result filters.

### DIFF
--- a/backend/test_observer/controllers/test_results/filter_test_results.py
+++ b/backend/test_observer/controllers/test_results/filter_test_results.py
@@ -122,7 +122,6 @@ def build_query_filters_and_joins(
 
     if len(filters.execution_metadata) > 0:
         query_filters.append(filter_execution_metadata(filters.execution_metadata))
-        joins_needed.add("execution_metadata")
 
     if len(filters.issues) > 0:
         query_filters.append(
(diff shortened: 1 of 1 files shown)


backend/test_observer/controllers/test_results/filter_test_results.py
```diff
@@ -0,0 +1,186 @@
...
+    if "test_case" in joins_needed:
+        query = query.join(TestResult.test_case)
+
+    if "execution_metadata" in joins_needed:
+        query = query.join(TestExecution.execution_metadata)
+
+    return query
+
+
+def filter_test_results(query: Select, filters: TestResultSearchFilters) -> Select:
```
rpbritton, line 168: This is purely a refactor/moving things around so it can be reused by the bulk attach endpoint. Functionality for the `GET /test-results` endpoint should be exactly the same.

PR #395, opened 2025-09-26
[TO-161] Add permissions to issue modification endpoints

rpbritton:
## Description

Getting my hands dirty with permissions by adding it to the issues endpoint. Since these aren't used by anyone yet I think this is a good safety feature.

Side note: I think the move for [TO-182](https://warthogs.atlassian.net/browse/TO-182) will be to have a config option that allows you to specify a set of base permissions that everyone has (including authenticated users).

We can then add permissions for ALL the endpoints, and modify the certification test observer to allow for all those base permissions (including modification endpoints), and that should allow tests.ubuntu.com to also use this and just specify the view only permissions.

## Resolved issues

[TO-161](https://warthogs.atlassian.net/browse/TO-161)


backend/test_observer/controllers/issues/issue_attachments.py
```diff
@@ -125,11 +132,26 @@ def modify_issue_attachments(
     return issue
 
 
+def require_bulk_permission(
```
rpbritton, line 135: Fancy permission.

uyarahmet (approved):
LGTM!

PR #401, opened 2025-10-07
[TO-182] Add permission to all API endpoints

rpbritton:
## Description

Adds permissions to all endpoints (except for those listed in `backend/scripts/check_endpoint_permissions.sh`).

Specific permissions can be disabled with the charm's `ignore_permissions` config option.

These changes are not backwards compatible. We need to configure the charm with ignore permissions before (while?) upgrading. We also need to give @andersson1234 a heads up.

## Resolved issues

[TO-182](https://warthogs.atlassian.net/browse/TO-182)

## Documentation

Charm config option includes description.

## Web service API changes

OpenAPI is up to date and includes the scopes necessary to call any routes.

## Tests

*All* unit tests updated to verify permission scopes of the endpoints they call.

Additionally a new pull request check added (`backend/scripts/check_endpoint_permissions.sh`) to ensure that all endpoints have permissions assigned to them.

[TO-182]: https://warthogs.atlassian.net/browse/TO-182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/terraform/ps5/test-observer.tf
+++ b/terraform/ps5/test-observer.tf
@@ -85,6 +85,11 @@ variable "sessions_secret" {
   type        = string
 }
 
+variable "ignore_permissions" {
+  description = "List of API permissions to ignore for all requests"
+  type        = list(string)
+}
+
 
 
 locals {
@@ -158,6 +163,7 @@ resource "juju_application" "test-observer-api" {
     saml_sp_cert          = var.saml_sp_cert
     saml_sp_key           = var.saml_sp_key
     sessions_secret       = var.sessions_secret
+    ignore_permissions    = join(",", var.ignore_permissions)
   }
 
   units = 3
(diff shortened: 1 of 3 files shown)


backend/charm/config.yaml
```diff
@@ -27,3 +27,7 @@ options:
   sessions_secret:
     description: Randomly generated secret key to use for signing session cookies
     type: string
+  ignore_permissions:
```
rpbritton, line 30: This is how we'll migrate. Just ignore all added permissions.

backend/test_observer/controllers/docs/docs.py
```diff
@@ -14,14 +14,45 @@
...
+from fastapi import APIRouter, Request
+from fastapi.responses import HTMLResponse, JSONResponse
 from fastapi.openapi.docs import get_swagger_ui_html
 
 
 router: APIRouter = APIRouter()
 
 
+@router.get("/openapi.json", include_in_schema=False)
+async def custom_openapi(request: Request):
```
rpbritton, line 26: Since we are using SAML and our own custom permissions scheme, the scopes required for certain endpoints do not show up by default in the docs. This adds them so they are viewable on /docs.

PR #450, opened 2025-11-16
Fix percent encoding in test results

uyarahmet:
## Description

When `execution_metadata` query parameters contain values with special characters (spaces, colons), the URL parameters were being **double URL-encoded**, resulting in malformed query strings like:
- `%2520` instead of `%20` (double-encoded space)
- `%253A` instead of `%3A` (double-encoded colon)

We decided to change the encoding logic to base64

## Resolved issues

[TO-195](https://warthogs.atlassian.net/browse/TO-195?atlOrigin=eyJpIjoiMDM4OTQ2ZjVjNzZlNGNjN2FhOWYwYmE0M2U3NThjNTMiLCJwIjoiaiJ9)

image

## Web service API changes

- `parse_execution_metadata()`: Base64 decode parameters, then extract category and value

## Tests

[TO-195]: https://warthogs.atlassian.net/browse/TO-195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontend/lib/models/execution_metadata.dart
```diff
@@ -78,10 +82,7 @@ abstract class ExecutionMetadata with _$ExecutionMetadata {
 
   List<String> toQueryParams() {
     return toRows()
-        .map(
-          (row) =>
-              '${Uri.encodeComponent(row.$1)}:${Uri.encodeComponent(row.$2)}',
-        )
+        .map((row) => base64Encode(utf8.encode('${row.$1}:${row.$2}')))
```
rpbritton: You must encode/decode category and value separately. Otherwise with category `charm_qa:charm` and value `kratos`, you get `charm_qa:charm:kratos`, but when you split it it turns into category `charm_qa` and value `charm:kratos`.

PR #452, opened 2025-11-17
Add python import sorting and update copyright headers

rpbritton:
## Description

I've wanted to enable import sorting for `ruff` for a while now. This PR does that, and runs the formatter.

See `pyproject.toml` for important changes.

Also updates copyright headers to 2026 so the license check passes.

## Resolved issues

Imports are not sorted.

## Documentation

N/A

## Web service API changes

N/A

## Tests

N/A


Hyask:
@rpbritton It'd be lovely to rebase that, re-run `ruff`, and merge it straight away to put that behind you, and help everyone not care about formatting.

wctaylor (approved):
Since I saw all the migrations get formatted, we could update `backend/migrations/script.py.mako` to match what `ruff` does, so then generated migrations will match. But it also doesn't really matter, since they will just get run through `ruff` anyway

PR #467, opened 2025-11-20
fix: treat all-skipped test executions as failures

moisesbenzan:
## Description

- Modify has_failures property to return True when all test results are skipped
- Add test coverage for test execution with all skipped results
- Ensure test execution status is set to FAILED when all tests are skipped

## Resolved issues

TS-62

## Tests

Tested locally


rpbritton (commented):
Ahh shoot I think this should be a logic change on charm-integration-testing with how we mark the test execution. I don't really want to change the existing functionality given that cert might depend on this behavior. wdyt?

PR #468, opened 2025-11-20
[TO-223] Add assignee filter to backend

rpbritton:
## Description

Adds assignee filter to test results. Also allows for searching the users endpoint.

## Resolved issues

[TO-223](https://warthogs.atlassian.net/browse/TO-223)

## Web service API changes

Users endpoint response format is modified. This endpoint is not currently used for anything so I am okay with this (should not break anything).

## Tests

Unit tests added.

[TO-223]: https://warthogs.atlassian.net/browse/TO-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/schemata/openapi.json
+++ b/backend/schemata/openapi.json
@@ -2566,7 +2566,15 @@
                 {
                   "type": "array",
                   "items": {
-                    "$ref": "#/components/schemas/QueryValue"
+                    "const": "any",
+                    "type": "string"
+                  }
+                },
+                {
+                  "type": "array",
+                  "items": {
+                    "const": "none",
+                    "type": "string"
                   }
                 },
                 {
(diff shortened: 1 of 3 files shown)


wctaylor (commented):
There seem to be two slightly different schemata for `assignee_ids` (and I think also for `issues`. 

1. It is one of `[int]`, `["any"]`, `["none"]`, or `null`
    * This is what is used in the `assignee_ids` parameter of `v1/test-results`
    * It also seems to be the assumption in the throughout `test_results/test_results.py`
2. It is one of `[int]`, `"any", `"none"`
    * This seems to be the assumption in `filter_test_results.py`, `test_results/models.py`
    * That is, `"any"` and `"none"` are not inside a list

Should the two be consistent? If so, which one is the desired one?

PR #470, opened 2025-11-21
Update Docs Starter Pack to 1.3.0

rkratky:
## Description

Updates the Starter Pack to latest version, which includes:

* Update build flow
* Remove old files
* Add CLA check

## Tests

- Docs checks pass with the exception of `make vale` (style-guide check), which is not blocking and can be fixed later.
- GH Actions docs workflows also updated.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -155,11 +155,6 @@ html_context = {
 
 html_extra_path = []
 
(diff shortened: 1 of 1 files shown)


tang-mm (approved):
Thank you for submitting the changes @rkratky ! 

@mariaclara-sia I left a few comments related to project-specific setup. Please take a look and let me know if you have questions

PR #483, opened 2025-11-27
Fix failing mypy checks

almeidaraul:
## Description

The mypy checks for the backend always give warnings due to 3 regex strings without proper escape characters.

This PR solves the issue by making them into raw strings.

## Tests

Tested with `pytest` to ensure that no tests started to fail.

### DIFF
--- a/backend/scripts/show_testcase_history.py
+++ b/backend/scripts/show_testcase_history.py
@@ -71,7 +71,7 @@ def get_previous_jenkins_submission_json_urls(url: str, count: int) -> list[str]
 
 
 def extract_build_number_from_jenkins_submission_json_url(url: str) -> int:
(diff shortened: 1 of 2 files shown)


wctaylor (commented):
Where do you see the missing import errors? If I run locally, I just see the invalid escape sequence messages, and if I look at a random other PR, like [this one](https://github.com/canonical/test_observer/actions/runs/19543726319/job/55956900920?pr=468), I also only see the invalid escape sequence messages.

Also, on a subjective style note, I tend to prefer raw strings for regex patterns, just to avoid all the extra slashes, e.g `pattern = r"(.+/)\d+/artifact/artifacts/submission\.json"` instead of `pattern = "(.+/)\\d+/artifact/artifacts/submission\\.json"`, but that's purely subjective

Issue #472, opened 2025-11-21
Add filter to exclude test runs that have been reran


rpbritton:
From both search test results page and artefacts page.

PR #492, opened 2025-12-02
[TO-239] Add rerun and latest filters for executions

rpbritton:
## Description

Adds filters for test results from executions that are:
* The latest in their test plan / artefact build / environment group
* Have a rerun requested

image

## Resolved issues

#472

## Documentation

N/A

## Web service API changes

OpenAPI shows the two new filters.

## Tests

Tested locally and with unit tests.


backend/test_observer/controllers/test_results/models.py
```diff
@@ -76,6 +76,8 @@ class TestResultSearchFilters(BaseModel):
     assignee_ids: list[int] | Literal[QueryValue.ANY, QueryValue.NONE] = Field(
         default_factory=list
     )
+    rerun_is_requested: bool | None = None
```
almeidaraul: nit: use Optional[bool] instead

rpbritton: While they're equivalent from my understanding, I prefer `bool | None`, and I believe that it is more commonly preferred in the broader python community (see https://stackoverflow.com/questions/69440494/python-3-10-optionaltype-or-type-none).

PR #513, opened 2026-01-07
feat: terraform: include service-hostname in nginx config

Hyask:
This should allow proper certificates to be generated through LEGO.


motjuste (approved):
Thanks! LGTM.

PR #518, opened 2026-01-12
Moved Flutter sources to "src" subdir and reworked Dockerfile

Vultaire:
## Description

This is a set of fairly small changes that are intended to improve frontend builds and provide a base for other potential improvements.

* The Dockerfile's "apt-get update" and "apt-get install" lines are now linked together as a joint step.  This avoids a podman/docker caching issue where an update to dependencies may require re-running the "apt-get install" line, but the preceding "apt-get update" line isn't reran as well, resulting in apt trying to pull older versions of packages which no longer exist in the repositories.
* The Flutter sources and most related files are moved from frontend/ to frontend/src/.  This is to allow for updates to the Dockerfile or nginx.conf file without retriggering large rebuilds unnecessarily, as the current Dockerfile ends up copying the full contents of the frontend/ directory and, even if a non-flutter-related file were to change, this would trigger a rebuild and waste time.

To be clear: yes, I would like to see us move away from Flutter for the UI as I see it having numerous weaknesses, primarily in relation to rendering to canvas rather than the DOM, and this minor restructuring makes it easier for us to do such experimentation.  But even if this were not to happen, I believe this PR's changes still offer some benefit.

## Related issues (not "resolved issues" in this case)

* [Test Observer experimental non-flutter rewrite work](https://warthogs.atlassian.net/browse/IQA-3353)

## Documentation

No meaningful changes.  Some README and similar files which were specific to the Flutter codebase were moved from frontend/ to frontend/src/ as well.

## Web service API changes

No changes.

## Tests

I did walk through the README at frontend/README.md (now frontend/src/README.md) and ran unit/integration tests and benchmarks - however one of the integration tests fails and the benchmarks also fail.  This being said, the current mainline also fails in the same way; I've observed no regressions compared to the mainline.

Of note though is I have not looked at how this is built for production nor how this is charmed, so I don't know if there may be some impact there.  As long as those processes are dealing with built docker containers, they should not be impacted as the directory structure change is handled via frontend/Dockerfile and should not have any meaningful impact on the final generated image.


Vultaire:
Dropped in favor of an alternate implementation.

PR #533, opened 2026-01-30
Update the license year to 2026

moisesbenzan:
## Description
Copyright checker was stuck at 2023, moved it to 2026.

## Resolved issues
No specific issue, just an annoyance 

## Documentation

No change in docs

## Web service API changes
No change

## Tests
CI tests run and pass


moisesbenzan:
> Seems reasonable. Do we need to bulk update every license header to 2026 now?

Yup, we should probably do it on its own PR.

PR #555, opened 2026-02-10
implement jira issue sync with service account tokens

uyarahmet:
## Description

This PR implements automated Jira issue synchronization for Test Observer using Atlassian's scoped service account tokens with cloud ID-based authentication.

**JiraClient**: New client uses scoped service account tokens
  - Authenticates via `api.atlassian.com/ex/jira/{cloud_id}` gateway
  - Uses HTTPBasicAuth with email + API token
  - Supports Jira REST API v3
 
**Factory Updates**: 
  - Jira service now requires `JIRA_CLOUD_ID`, `JIRA_EMAIL`, `JIRA_API_TOKEN`

### DIFF
--- a/backend/test_observer/external_apis/jira/jira_client.py
+++ b/backend/test_observer/external_apis/jira/jira_client.py
@@ -74,7 +74,7 @@ class JiraClient:
             data = response.json()
 
             fields = data.get("fields", {})
-            status = (fields.get("status") or {}).get("name", "Unknown")
+            status = fields.get("status", {}).get("name", "Unknown")
 
             return IssueData(
                 title=fields.get("summary", ""),
--- a/backend/tests/external_apis/test_external_api_clients.py
+++ b/backend/tests/external_apis/test_external_api_clients.py
@@ -39,7 +39,7 @@ class TestJiraClient:
         assert client.api_token == "test-token"
         assert client.timeout == 30
 
-    def test_init_bearer_auth(self) -> None:
+    def test_init_with_custom_timeout(self) -> None:
         """Test JiraClient with custom timeout"""
         client = JiraClient(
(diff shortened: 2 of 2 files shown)


backend/test_observer/external_apis/jira/jira_client.py
```diff
@@ -14,131 +14,78 @@
...
+                auth=HTTPBasicAuth(self.email, self.api_token),
+                headers={"Accept": "application/json"},
+                timeout=self.timeout,
             )
+
+            response.raise_for_status()
+            data = response.json()
+
+            fields = data.get("fields", {})
+            status = (fields.get("status") or {}).get("name", "Unknown")
```
wctaylor: [nitpick]

```suggestion
            status = fields.get("status", {}).get("name", "Unknown")
```

backend/tests/external_apis/test_external_api_clients_integration.py
```diff
@@ -24,22 +27,48 @@
...
+            api_token=cast(str, os.getenv("JIRA_API_TOKEN")),
+        )
+        result = client.get_issue("TO", "TO-169")
 
         assert result.title is not None
-        assert result.state in ("open", "closed")
+        assert result.state is not None
         assert result.raw is not None
 
+    @pytest.mark.skipif(
```
almeidaraul: why not require that the credentials are configured instead of skipping the tests?

wctaylor: Raul, I think you raise a good point, and it would probably be good to run tests with access to the credentials, but since this is a public repo, I want to spend some time better understanding the implications, and what sort of hardening might be required of the workflows when passing secrets in this case.

I think we can implement the use of those credentials in a separate PR

PR #570, opened 2026-02-16
Temporarily disable issue synchronization

wctaylor:
## Description

Following a production deployment on February 12, Test Observer is very slow and database connections are getting exhausted while queries sit idle in transactions. The oldest query has been idle for 16 hours at this point, and it is related to the issue synchronizer. The logs for the API pod are also filled with `No synchronizer available for issue` warnings. 

Non-synchronizer queries are also stuck, but since these two things jump out as synchronizer-related, I think the first step is to disable the synchronizer and see if things improve. If nothing else, this should cut down on log noise and help find the real culprit.

Using `git revert` was getting confusing, and we don't necessarily need to remove all the associated code and tests, we just need to not call it in production, so commenting it out seemed simplest for now.

The backend test fails because there is commented out code, so if we don't want to override that, I can try to properly deal with the reverts.

### DIFF
--- a/.github/workflows/test_backend.yml
+++ b/.github/workflows/test_backend.yml
@@ -24,8 +24,8 @@ jobs:
       - run: pip install uv
       - run: uv sync --all-groups
       - run: uv pip install -e .
-      - run: uv run ruff check
-      - run: uv run ruff format --check
+#      - run: uv run ruff check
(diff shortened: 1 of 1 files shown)


rpbritton (approved):
Sorry I meant to make them pass, but I guess if this is temporary then fine.

PR #575, opened 2026-02-18
[Fix] Get or create should not commit

rpbritton:
## Description

The `get_or_create` function should not commit because it is used in some functions (like start test), that should really be a transaction.

## Resolved issues

Environment review exists without a test execution meaning ghost tests.

## Documentation

Function change documented

## Web service API changes

N/A

## Tests

Existing tests pass. Asked copilot to check for use cases where this feature was being relied on.

### DIFF
--- a/backend/tests/data_access/test_repository.py
+++ b/backend/tests/data_access/test_repository.py
@@ -157,3 +157,36 @@ def test_get_or_create_race_condition(
 
     assert result.id == test_case.id
     assert calls["one"] == 1
+
+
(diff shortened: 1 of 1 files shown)


motjuste (commented):
Aight ...

I noticed that the `commit` was there before the last change too ...  I am not clear why it was not a problem before ... however ... this kinda makes sense ...

I also confirmed that we have a `db.commit()` after each use of this function.

We should add it to the documentation and have a test too.  Would be super easy changes.

PR #579, opened 2026-02-19
Reduce connection pool size

wctaylor:
## Description

The current connection pool size looks targeted for a single unit, but is capable of saturating the default PostgreSQL connection limit of 100 when there are multiple units. With a single unit, it caps at 90 connections, leaving 10 connections reserved. But with three units, where each unit can create up 90 connections, and will happily create and leave up to 45 connections open each, we can attempt to create up to 135 idle connections (and 270 maximum connections), which Postgres will kill with complaints about no available connections.

A pool size of 10 with overflow of 20 will cap at 30 connections, so with three workers, we return to 90 max connections with 10 reserved.

### DIFF
--- a/backend/test_observer/data_access/setup.py
+++ b/backend/test_observer/data_access/setup.py
@@ -29,7 +29,7 @@ from sqlalchemy.orm import sessionmaker
 DEFAULT_DB_URL = "postgresql+pg8000://postgres:password@test-observer-db:5432/postgres"
 DB_URL = environ.get("DB_URL", DEFAULT_DB_URL)
 
-engine = create_engine(DB_URL, pool_size=45, max_overflow=45)
+engine = create_engine(DB_URL, pool_size=10, max_overflow=20)
 SessionLocal = sessionmaker(
     autocommit=False,
     autoflush=False,


motjuste (approved):
I am approving this as I find it reasonable as is, but do look at https://github.com/canonical/test_observer/pull/579#issuecomment-3926997717 ... 

`pool_recycle` may not be that useful for `pg8000` driver (per CLAUDE) ... but still may catch edge cases ...

It also suggests ... in case where connection to database may get flaky ... to set `pool_pre_ping=True`

All this can still be done in a separate PR ... so approving

wctaylor:
Based on the context from the [other pool increase PR](https://github.com/canonical/test_observer/pull/197) (before the one you link), it sounds like they were indeed only running one unit at that time, since the suggestion was that running with up to 50 connections would allow them to add another unit without breaking the 100 connection limit.

I think things are in a good enough state for now that we don't need to use `pool_recycle` or `pool_pre_ping` (I don't think we saw any instances of flaky connections - just connections getting stuck in transactions, and those are now getting killed automatically by Postgres if they are stuck too long)

PR #569, opened 2026-02-13
Feat: Allow frontend tabs customization with config.yaml

mclemenceau:
**AI Use Disclosure**: This change was authored with assistance from AI. The code was reviewed, tested locally, and validated as much as possible.

## Description

This pull request introduces a configurable frontend navigation system via a YAML configuration file. It adds support for Testing Tabs configuration both in the backend charm and the Flutter frontend. Only valid tab names are accepted. The changes also include unit tests for the new configuration parsing logic.
With Tab customization you can now:
 - Display only a selected choice of Tabs
 - Reorder Testing Tabs based on your test_observer workflow

Examples:

with frontend/assets/config.yaml
``` yaml
tabs:
  - images
```
Screenshot from 2026-02-17 08-41-25

with frontend/assets/config.yaml
``` yaml
tabs:
  - images
  - snaps
```
Screenshot from 2026-02-17 08-37-57

**Frontend tab configuration support:**

* Added a new `frontend-config` charm config option to allow specifying frontend tab configuration as a YAML string;
* Introduced a default `config.yaml` in the frontend assets and logic to read and parse this file at app startup, setting the list and order of tabs dynamically. 

**Frontend UI and routing updates:**

* Updated navigation bar and router logic to use the dynamically configured tab list, replacing hardcoded tab routes and names with values from the loaded configuration.
* Added a helper function to map route names to human-friendly tab display names.

**Testing and validation:**

* Added a new test suite for the frontend configuration parser, covering valid and invalid YAML inputs, tab values, and edge cases.

## Resolved issues

This resolves issue #548 

## Documentation
 - minimal config.yaml syntax was included in frontend-config.dart
 - 
## Web service API changes

No API Changes

## Tests

- Local frontend test are passing
- Fixed linting according to GitHub Actions results
- Performed various manual integration testings
- Not able to test charm at the moment and faving issue with Github action charm integration which maybe related to permission testing charm with Github Actions

PR #581, opened 2026-02-19
terraform: update frontend-config for tests.ubuntu.com

Hyask:
See #569

### DIFF
--- a/terraform/ps6/variables.tf
+++ b/terraform/ps6/variables.tf
@@ -83,6 +83,12 @@ variable "frontend_hostname" {
   type        = string
 }
 
+variable "frontend_config" {
+  description = "config for frontend"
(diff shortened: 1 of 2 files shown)


terraform/ps6/variables.tf
```diff
@@ -83,6 +83,16 @@ variable "frontend_hostname" {
   type        = string
 }
 
+variable "frontend_config" {
+  description = "config for frontend"
+  type        = string
+  default     = <<-EOT
```
rpbritton: I don't think this default should be here. If this variable is unset, shouldn't it just use the charm's default?

You can set this variable when you init the module in the other code.

Also I'd recommend using something like `file("${path.module}/config.yaml")` instead of inline yaml.

PR #587, opened 2026-02-24
TO-35: add artefact matching rules

almeidaraul:
## Description

This pull request introduces a new "Artefact Matching Rule" feature, which will replace the "reviewer_families" mechanism for associating artefact families with teams. This will allow for more complex reviewer assignment.

Essentially, the Artefact Matching Rule is an entity associated with teams. In a subsequent PR, during reviewer assignment, an artefact will be matched against a rule, and then the reviewer will be selected from the teams associated to the rule.

## Resolved issues

Partially resolves TO-35


backend/migrations/versions/2026_02_27_1246-40b42f906881_add_artefact_matching_rules.py
```diff
@@ -0,0 +1,50 @@
+"""add artefact matching rules
```
rpbritton: No copyright header

rpbritton (changes requested):
I don't really have much in the way of comments. Seems like this particular PR is pretty low risk (just adding stuff, not really replacing or actually using it yet). Straight forward backend API seem inline with existing structures.

backend/test_observer/controllers/common/__init__.py
rpbritton: I think this can be removed.

Issue #553, opened 2026-02-09
API: endpoints supporting / requiring pagination must contain the count, offset, and limit used in response


motjuste:
Currently we have a few endpoints that support paginated output, e.g. `GET /v1/test-results`, `GET /v1/issues`, and so on.  These endpoints seem to have a default `limit` and `offset` parameters set when not provided by the caller, making pagination mandatory.

Some **but not all** of these respond with the total `count` of results expected from the endpoint.  None of them AFAICT respond with `offset` or `limit` used in the response.

Having `count`, `offset`, and `limit` in such **mandatorily** paginated endpoints helps consumers of the API to more easily keep track of the progress the caller has made through these paginated responses: based on the three values in the response, the caller would know if they need to make subsequent requests, **especially** since the original URL for the request may not have included any `limit` or `offset` param, and still be paginated due to default parameters.

PR #590, opened 2026-02-25
[TO-288] - Include count, offset, and limit in all paginated responses

uyarahmet:
## Description

Closes https://github.com/canonical/test_observer/issues/553

All endpoints that support pagination now return count, limit, and offset alongside their data. This lets API consumers determine total result size, know how many pages remain, and understand what pagination parameters were actually applied (including defaults when none were passed).
## Resolved issues

[TO-288](https://warthogs.atlassian.net/browse/TO-288)
## Web service API changes

All paginated endpoints now include three additional fields in their response body:

- count — total number of records matching the current filters, regardless of pagination
- limit — the maximum number of results used for this page (echoes the default of 50 if not passed)
- offset — the number of results skipped for this page (echoes 0 if not passed)

Example response shape for GET /v1/issues:
```
{
  "issues": [...],
  "count": 142,
  "limit": 50,
  "offset": 0
}
```

### Count Query Approach:
For the three endpoints that needed a new count query (/artefacts/search, /environments, /issues), we wrap the filtered query as a subquery: select(func.count()).select_from(query.subquery()). This keeps the code DRY by avoiding duplication of filter logic, and the performance overhead is negligible for these endpoints.

## Tests

All controllers' pytests are updated following these changes.

[TO-288]: https://warthogs.atlassian.net/browse/TO-288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/tests/controllers/artefacts/test_artefacts_search.py
+++ b/backend/tests/controllers/artefacts/test_artefacts_search.py
@@ -250,3 +250,26 @@ def test_search_artefacts_default_limit(
     artefacts = response.json()["artefacts"]
     assert len(artefacts) <= 50
     assert len(artefacts) == 50
+
+
+def test_search_artefacts_pagination_metadata(
+    test_client: TestClient, generator: DataGenerator
+):
+    """count reflects total results, limit and offset echo back the used values"""
+    unique_marker = uuid.uuid4().hex[:8]
+    for i in range(5):
+        generator.gen_artefact(name=f"meta_artefact_{unique_marker}_{i:02d}")
+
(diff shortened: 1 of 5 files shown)


motjuste (approved):
LGTM as is ... Thanks a lot for the feature and the tests!

I hope this doesn't break the frontend in any way ... please verify that it doesn't.

I also didn't find a solution better than an extra query to get the total counts.

PR #597, opened 2026-02-27
TO-51: assign multiple reviewers to large artefacts

almeidaraul:
Assign multiple reviewers to artefacts with more than 50 environments

Part of TO-51

Unit tests added


rpbritton (approved):
I think this will be good enough for now. Probably worth mentioning this feature exists in the TO channel after deploying it.

PR #602, opened 2026-02-28
Config cleanup

wctaylor:
## Description

This cleans up a few config files (mostly VS Code) and adds the `docs` as another root in the VS Code multi-root workspace, since it is also essentially it's own project and does have it's own Python environment.


rpbritton (approved):
lgtm

Issue #526, opened 2026-01-19
Showing issues not attached to any runs in the issues page


motjuste:
We have multiple issues which have not been attached to any runs, but still show up (and IMO clutter) the [issues page](https://test-observer.canonical.com/#/issues), e.g.: [131](https://test-observer.canonical.com/#/issues/349), [296](https://test-observer.canonical.com/#/issues/365), [291](https://test-observer.canonical.com/#/issues/364).

Many of such issues actually may not be relevant to be attached to specific test runs, so it may not be useful to show them up in the issues page.

- We should show number of effected runs alongside the issue item of the table in the main issues page.
- It should be possible to filter them out at least by: "has runs" to which it is attached.

PR #608, opened 2026-03-04
[TO-271] - Show test results count per issue on issues page

uyarahmet:
## Description

Closes https://github.com/canonical/test_observer/issues/526

Each issue in `GET /v1/issues` now includes an `test_executions_count` field, which is the number of distinct test executions attached to that issue. This is displayed alongside the issue status on the issues list page.

The count is computed via a correlated subquery (`COUNT(DISTINCT test_execution_id)`) at query time.  No schema changes required. Note this counts runs (test executions), not individual test results, which is the more meaningful metric: e.g. an issue affecting 4 test cases in the same run counts as 1 affected run.

## Resolved issues

[TO-271](https://warthogs.atlassian.net/browse/TO-271)
## Web service API changes

- `GET /v1/issues` response now includes `test_executions_count: int` on each issue object (defaults to 0 for issues with no attachments)

## Tests

Added unit tests and updated relevant tests.

image

image

[TO-271]: https://warthogs.atlassian.net/browse/TO-271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/schemata/openapi.json
+++ b/backend/schemata/openapi.json
@@ -5274,9 +5274,9 @@
             ],
             "title": "Labels"
           },
-          "test_results_count": {
+          "test_executions_count": {
             "type": "integer",
-            "title": "Test Results Count",
+            "title": "Test Executions Count",
             "default": 0
           }
         },
--- a/backend/test_observer/controllers/issues/issues.py
+++ b/backend/test_observer/controllers/issues/issues.py
@@ -157,8 +157,10 @@ def get_issues(
     rows = db.execute(stmt.add_columns(runs_count_subq)).all()
     return IssuesGetResponse(
         issues=[
-            MinimalIssueResponse.model_validate(issue).model_copy(update={"test_results_count": test_results_count})
-            for issue, test_results_count in rows
+            MinimalIssueResponse.model_validate(issue).model_copy(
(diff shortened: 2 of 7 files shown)


uyarahmet:
> That looks beautiful and useful as is ...
> 
> I know that in the original issue there was also a request to implement filtering the issues on the issue page based on these counts (at least, remove issues with 0 affected runs) ... but what are the challenges to implement that?

Filtering is possible but I'd suggest we hold off for now. Issues with 0 affected runs aren't necessarily noise, they may be newly created and not yet attached to any test results, or they may be something like test observer [issues](https://github.com/canonical/test_observer/issues) which aren't supposed to be attached to test results.

backend/schemata/openapi.json
```diff
@@ -5273,6 +5273,11 @@
               }
             ],
             "title": "Labels"
+          },
+          "affected_runs_count": {
```
rpbritton: I think the implementation here looks good, assuming that it is fast. My problem is a naming nitpick though, we don't use `run` for anything. We have `test executions` and `test results`, so maybe rename this just `test_results_count`?

PR #644, opened 2026-03-12
Use PostgreSQL 14 in docker-compose

wctaylor:
## Description

The backend charm deploys alongside PostgreSQL 14. Recent changes introduced a migration which relied on a feature introduced in PostgreSQL 15, which passed CI tests using PostgreSQL 15 but broke when attempted by the charm.

This PR uses PostgreSQL 14 in the `docker-compose.yml` file to match the current expectations of the backend charm and removes the offending part of the migration and model definition.

## Resolved issues

CI should be able to pass again. We should also hopefully catch future issues like this sooner.

### DIFF
--- a/backend/schemata/openapi.json
+++ b/backend/schemata/openapi.json
@@ -4609,37 +4609,19 @@
             "$ref": "#/components/schemas/FamilyName"
           },
           "stage": {
-            "anyOf": [
-              {
-                "type": "string"
-              },
-              {
-                "type": "null"
-              }
-            ],
-            "title": "Stage"
+            "type": "string",
+            "title": "Stage",
+            "default": ""
           },
           "track": {
-            "anyOf": [
-              {
-                "type": "string"
-              },
-              {
-                "type": "null"
-              }
-            ],
-            "title": "Track"
+            "type": "string",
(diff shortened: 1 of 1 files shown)


backend/test_observer/data_access/models.py
```diff
@@ -207,7 +207,7 @@ class ArtefactMatchingRule(Base):
 
     __table_args__ = (
         UniqueConstraint(
-            "family", "stage", "track", "branch", postgresql_nulls_not_distinct=True
```
almeidaraul: by removing this, ("snap", null, null, null) will not be the same as ("snap", null, null, null) in the database. we can write an application-level check but it's not as safe as enforcing it from the database side

wctaylor: We could also do something like this I think

```
__table_args__ = (
        Index(
            "uq_rule_logic_null_safe",
            "family",
            func.coalesce(stage, '-'),
            func.coalesce(track, '-'),
            func.coalesce(branch, '-'),
            unique=True,
        ),
    )
```

wctaylor: I like empty strings better, though

Issue #609, opened 2026-03-04
Fine grained team ACLs


Hyask:
The current ACL system is nice, but a little bit limited.

One great improvement that we would need is the ability to give some permissions to some teams, and those would apply only to the artifacts that the team owns.

Let's take the following example covering precisely our use-case:

Image | Owner
---|---
ubuntu.iso  | canonical-desktop-team
xubuntu.iso | xubuntu-release

In that case with both teams having `change_artefact` and `change_rerun`, each of them should be able to set their artifact as ready, and trigger reruns, without being able to do it on the image they don't own.

Regarding implementation, I've no precise idea how the system works, but it would be nice to still have `change_artefact` and `change_rerun` to keep behaving as they do now, meaning being able to manage all artefacts, because that would allow us to have a team of admins. Introducing a more reduced permission set, like `change_owned_artefact` and `change_owned_rerun`, to implement the required permissions, would totally work for us.

I hope everything makes sense.

PR #660, opened 2026-03-24
TO-299: Add 'name' field to AMRs

almeidaraul:
## Description

Adds the `name` field to `ArtefactMatchingRules`. This will be useful for controlling permissions for specific artefacts in #609 

## Resolved issues

Part of TO-299

### DIFF
--- a/.github/.jira_sync_config.yaml
+++ b/.github/.jira_sync_config.yaml
@@ -46,7 +46,7 @@ settings:
   sync_comments: true
   
   # (Optional) (Default: None) Parent Epic key to link the issue to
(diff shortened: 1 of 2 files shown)


backend/test_observer/data_access/models.py
```diff
@@ -209,6 +209,7 @@ class ArtefactMatchingRule(Base):
 
     __tablename__ = "artefact_matching_rule"
 
+    name: Mapped[str | None] = mapped_column(String(200), unique=True, nullable=True)
```
rpbritton: Shouldn't this be a string to satisfy the unique constraint?

PR #674, opened 2026-03-31
Move additional endpoints behind authentication

wctaylor:
## Description

This PR moves a few extra paths behind authentication to give more control over what can be viewed and by whom. This is less about hiding information (it's a public repo after all) and more about providing consistent handling in the running application. It introduces a `REQUIRE_AUTHENTICATION` config option, which will allow private deployments to put some endpoints behind authentication and permissions, while enabling public deployments to remain publicly visible.

It also removes explicit notification permissions, as being authenticated is taken as sufficient to view notifications.
There are still tests confirming that one user cannot view or change another user's notifications.

## Resolved issues

Fixes: [Test Observer: Move root and docs behind authentication](https://warthogs.atlassian.net/browse/IQA-3506)

## Documentation

The `openapi.json` file is updated.

## Web service API changes

The following endpoints can optionally require authentication (no permission is needed beyond being authenticated)

* `/`
* `/docs`
* `/openapi.json`
* `/v1/applications/me`
* `/v1/users/me`
* `/v1/users/{id}/notifications`
* `/v1/users/{id}/notifications/count`
* `/v1/users/{id}/notifications/{notification_id}/dismiss`
* `/v1/version`

A new `view_sentry_debug` permission has been added for the  `/sentry-debug` endpoint.

## Tests

Unit tests are added to cover the authentication scenarios and permission handling


rpbritton (dismissed):
This all looks reasonable to me.

wctaylor:
Since I originally started this PR, additional changes have been merged that move the `permissions` to a PostgreSQL enum. This turns out to be a bit of a headache, so I will do one or two things:

1. (Certainly) Draft a separate PR that only adds the new permissions (because this is now a headache), but does not use them anywhere, assuming that we stick with the PostgreSQL enum.
2. (Possibly) Draft a PR that undoes the PostgreSQL enum

backend/scripts/check_endpoint_permissions.sh
```diff
@@ -24,20 +24,13 @@ OPENAPI_JSON="$1"
...
   {"method": "get", "path": "/v1/auth/saml/logout"},
   {"method": "post", "path": "/v1/auth/saml/acs"},
   {"method": "get", "path": "/v1/auth/saml/sls"},
-  {"method": "post", "path": "/v1/auth/saml/sls"},
-  {"method": "get", "path": "/v1/users/me"},
-  {"method": "get", "path": "/v1/applications/me"},
-  {"method": "get", "path": "/health/live"},
-  {"method": "get", "path": "/health/ready"}
+  {"method": "post", "path": "/v1/auth/saml/sls"}
 ]'
```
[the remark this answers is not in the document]
wctaylor: Yes, for some reason that workflow stopped running, and the CI passed despite this needing to fail...but it is updated now

backend/scripts/check_endpoint_permissions.sh
```diff
@@ -23,21 +23,24 @@ OPENAPI_JSON="$1"
...
   {"method": "get", "path": "/v1/auth/saml/sls"},
   {"method": "post", "path": "/v1/auth/saml/sls"},
-  {"method": "get", "path": "/v1/users/me"},
+  {"method": "get", "path": "/"},
+  {"method": "get", "path": "/v1/version"},
   {"method": "get", "path": "/v1/applications/me"},
-  {"method": "get", "path": "/health/live"},
-  {"method": "get", "path": "/health/ready"}
+  {"method": "get", "path": "/v1/users/me"},
+  {"method": "get", "path": "/v1/users/{id}/notifications"},
```
rpbritton, line 42: Shouldn't this just be `/v1/users/me/notifications`?

backend/scripts/seed_data.py
```diff
@@ -799,9 +797,25 @@ def seed_data(client: TestClient | requests.Session, session: Session | None = N
...
+    # which would cause an error when trying to add the application.
     application = session.scalar(select(Application).where(Application.name == "seed_data_app"))
+    inspector = inspect(session.get_bind())
+    permissions = next(e["labels"] for e in inspector.get_enums() if e["name"] == "permission")  # type: ignore
     if not application:
-        application = Application(name="seed_data_app", permissions=[p.value for p in Permission])
+        # This is to allow CI to generate and provide an API key,
+        # which will allow for easy authentication in CI tests,
+        # such as comparing the live OpenAPI spec with the committed version.
+        if api_key := os.getenv("SEED_DATA_APP_KEY", ""):
```
rpbritton: I'm fine with this, but an alternative would just be a hard coded token. Since it is just seed data there shouldn't be any security concerns.

rpbritton (changes requested):

backend/scripts/seed_data.py
wctaylor: Yeah, I originally used a hardcoded token, but I let Copilot convince me otherwise, and I do still think it's ok

backend/scripts/check_endpoint_permissions.sh
wctaylor: The script errors if I do that

backend/test_observer/controllers/notifications/notifications.py
```diff
@@ -97,6 +96,11 @@ def get_notifications(
     db: Session = Depends(get_db),
 ):
     """Get all notifications for the specified user"""
+
+    # Notifications only make sense in the context of a user
+    if user is None:
+        raise HTTPException(status_code=401, detail="Not authenticated")
+
```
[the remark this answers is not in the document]
wctaylor: I think it's fine

backend/test_observer/controllers/notifications/notifications.py
```diff
@@ -134,6 +138,10 @@ def mark_notification_as_read(
     db: Session = Depends(get_db),
 ):
     """Mark a notification as read"""
+    # Notifications only make sense in the context of a user
+    if user is None:
+        raise HTTPException(status_code=401, detail="Not authenticated")
+
```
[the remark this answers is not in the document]
wctaylor: I think it's fine

backend/test_observer/controllers/router.py
```diff
@@ -61,13 +63,13 @@
 router.include_router(health.router, prefix="/health")
 
 
-@router.get("/")
+@router.get("/", dependencies=[Depends(authentication_checker)])
```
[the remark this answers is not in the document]
wctaylor: Honestly not really sure why this endpoint exists. I'm going to ignore this for now. Plus, browser users can access it through the Swagger UI

backend/test_observer/controllers/application/version.py
```diff
@@ -16,12 +16,14 @@
...
-from fastapi import APIRouter
+from fastapi import APIRouter, Depends
+
+from test_observer.common.permissions import authentication_checker
 
 router = APIRouter()
 
 
-@router.get("")
+@router.get("", dependencies=[Depends(authentication_checker)])
```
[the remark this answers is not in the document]
wctaylor: Browser users can access it through Swagger

backend/scripts/check_endpoint_permissions.sh
rpbritton: Doesn't this mean anyone can read/write anyone else's notifications though?

I feel like there should be `view_notification` and `change_notification` permissions that allow of that. Then in the endpoint it should check if `{id}` is `me` (or the authenticated user's id) and should bypass. Is that possible?

rpbritton (changes requested):

backend/scripts/check_endpoint_permissions.sh
wctaylor: The endpoint handling already finds the user through `target_user = _resolve_user_id(id, user, db)`. If it didn't, then presumably even having the `change_notification` permission would also be sufficient to change other people's notifications. 

And we do have a `test_mark_notification_as_read_wrong_user` that checks that users can't clear other users' notifications.

That said, I am happy to restore the permissions (and add a test to ensure users can't view other users notifications) if you would like

wctaylor: That said, it is also confusing. The implementation of `_resolve_user_id` actually suggests a user _could_ do stuff do another user, but I think we rely on it being called from the endpoint handling functions, which only pass the user returned by `get_current_user` to it. 

So we could also potentially remove the `_resolve_user_id` function entirely and change this

```
target_user = _resolve_user_id(id, user, db)
```

to this

```
if id != "me" and id != user.id:
    raise HTTPException(status_code=403, detail="Forbidden")
```

where `user` is still the input from `user: User | None = Depends(get_current_user)`

rpbritton: What behavior I think makes sense is:

- authenticated user can view/dismiss their own notifications
- user with `change_notification` can dismiss *other* users notifications

I think `test_mark_notification_as_read_wrong_user` would fail if ran with `test_client.post(f"/v1/users/{user.id}/notifications/{notification.id}/dismiss", headers={"X-CSRF-Token": "1"})`

wctaylor: Ah yes, ok, you're right about the test. I knew something was feeling wrong to me. I have updated the tests to include that test case.

I guess I still feel a little weird about users being able to dismiss other users notifications. Like, it seems that admins should be able to, in which case, we still wouldn't need the permission, but I don't know when we would have non-admin users with the `change_permission` permission. That said, it is at least consistent with other approaches, so I will go ahead and restore it

wctaylor:
Ok, this PR is getting pretty big and has evolved a decent bit. I'm going to try to split it up into smaller PRs, but I'll leave this open until that's done

PR #686, opened 2026-04-04
Add new permission values (PostgreSQL Enum)

wctaylor:
## Description

This PR introduces some new permissions, which will ultimately be used by https://github.com/canonical/test_observer/pull/674, but following the move to a PostgreSQL enum for permissions, it becomes more cumbersome to add permissions.

This PR represents what is necessary to add permissions in the current scheme. It can be merged if we decide to stick with it, but I think it also serves as a good indicator that we actually don't want to use a PostgreSQL enum for permissions going forward, and so we probably want to revert that.

## Resolved issues

Part of [Test Observer: Move root and docs behind authentication](https://warthogs.atlassian.net/browse/IQA-3506)

## Tests

I can build and run the Docker services locally, and I also executed the downgrade manually as a test. Both directions do seem to work. Checking with the `seed_data_app` that gets populated in the local database with every permission, I can see that it gets the new permissions on upgrade, and it successfully reverts to the old permissions on downgrade.

### DIFF
--- a/backend/migrations/versions/2026_03_31_1410-f0847700d32e_convert_permission_to_enum.py
+++ b/backend/migrations/versions/2026_03_31_1410-f0847700d32e_convert_permission_to_enum.py
@@ -81,6 +81,9 @@ def downgrade() -> None:
     # restore previous default
     op.execute("ALTER TABLE team ALTER COLUMN permissions SET DEFAULT '{}'::character varying[]")
 
+    # drop default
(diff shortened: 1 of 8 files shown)


rpbritton (changes requested):
Why not just use `ALTER TYPE enum_type ADD VALUE 'new_value';`?

https://stackoverflow.com/questions/1771543/adding-a-new-value-to-an-existing-enum-type/7834949#7834949

PR #692, opened 2026-04-06
TO-40: send notifications on assignment

almeidaraul:
## Description

This PR adds the functionality to actually send notifications when reviewers are assigned to artefacts and environments

## Resolved issues

Final part of TO-40


backend/test_observer/services/review_notification_service.py
```diff
@@ -0,0 +1,138 @@
+# Copyright 2024 Canonical Ltd.
```
rpbritton: 2024?

backend/test_observer/services/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2024 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License version 3, as
+# published by the Free Software Foundation.
+# This program is distributed in the hope that it will be useful,
```
rpbritton: Is this `services/` folder just meant to be shared utilities?

rpbritton (dismissed):
Seems straightforward and looks good to me.

backend/test_observer/services/review_notification_service.py
almeidaraul: wait it's 2026 already???

backend/test_observer/services/__init__.py
almeidaraul: yeah, do you think I should have this code somewhere else instead?

rpbritton: Maybe `backend/test_observer/common/`? You did create `backend/test_observer/common/helpers.py` last week.

almeidaraul: That makes sense, I wrote this in a separate directory just because it's a more complex structure than a single function or class but I'll move it to `common/` instead

backend/test_observer/common/review_notification.py
wctaylor, line 1: Why are the Jira cards created as part of the notification process, rather than as part of the assignment process?

backend/test_observer/common/review_notification.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@dataclass
+class BatchReviewerAssignedMessage:
+    """Context of a batch of reviewers being assigned to an artefact
+
+    Used for notifications and issue creation"""
+
+    artefact: Artefact
+    assigned_reviews: list[tuple[User, list[NotificationType]]]
```
amc94, line 42: nit: assigned_reviewers

backend/test_observer/common/review_notification.py
```diff
@@ -0,0 +1,131 @@
...
+        artefact: The artefact they were assigned to review
+        notification_type: The type of notification to create
+    """
+    for reviewer in reviewers:
+        try:
+            create_reviewer_notification(db, reviewer, artefact, notification_type)
+        except Exception:
+            logger.exception(
+                f"Failed to create {notification_type} notification for reviewer {reviewer.id} "
+                f"on artefact {artefact.id}"
```
amc94: why iterate with add + flush vs something like [add_all](https://docs.sqlalchemy.org/en/21/orm/session_api.html#sqlalchemy.orm.Session.add_all)

backend/test_observer/controllers/test_executions/start_test.py
```diff
@@ -121,19 +135,31 @@ def _assign_reviewers_to_environments(self) -> None:
         reviews_per_reviewer = _ceil_division(len(env_reviews), len(self.artefact.reviewers))
 
         current_reviewer = 0
+        newly_assigned_environment_reviewers: dict[int, User] = {}
```
amc94: why can't this just be a set of users if you only ever use the .values() of the dict?

backend/test_observer/controllers/test_executions/start_test.py
```diff
@@ -154,10 +180,36 @@ def assign_reviewer(self):
...
+                            [
+                                (reviewer, [NotificationType.USER_ASSIGNED_ARTEFACT_REVIEW])
+                                for reviewer in newly_assigned_reviewers
+                            ]
+                            + [
+                                (reviewer, [NotificationType.USER_ASSIGNED_ENVIRONMENT_REVIEW])
+                                for reviewer in newly_assigned_environment_reviewers
+                            ]
+                        ),
+                    )
```
amc94, line 215:
```suggestion
                if self.artefact.jira_issue is not None:
                    artefact_reviewers = [
                                (reviewer, [NotificationType.USER_ASSIGNED_ARTEFACT_REVIEW])
                                for reviewer in newly_assigned_reviewers
                            ]
                    environment_reviewers = [
                                (reviewer, [NotificationType.USER_ASSIGNED_ENVIRONMENT_REVIEW])
                                for reviewer in newly_assigned_environment_reviewers
                            ]
                    return BatchReviewerAssignedMessage(
                        artefact=self.artefact,
                        assigned_reviews = artefact_reviewers + environment_reviewers

                       
                    )
```

amc94: obviously format it

amc94 (changes requested):
mainly the add_all vs add thing

backend/test_observer/common/review_notification.py
almeidaraul: I don't agree with this

It's an assigned review because there's a notification type and the user, it's not just the user

backend/test_observer/common/review_notification.py
```diff
@@ -0,0 +1,119 @@
...
+    Args:
+        db: Database session
+        reviewers: List of users who were newly assigned as reviewers
+        artefact: The artefact they were assigned to review
+        notification_type: The type of notification to create
+    """
+    notifications = []
+    for reviewer in reviewers:
+        try:
+            notification = _create_notification_for_reviewer(
```
amc94: this doesn't need to be in a try block does it?

backend/test_observer/common/review_notification.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def _create_notification_for_reviewer(
+    user_id: int, notification_type: NotificationType, target_url: str
+) -> Notification:
+    return Notification(
+        user_id=user_id,
+        notification_type=notification_type,
+        target_url=target_url,
+    )
```
wctaylor: Does this function provide any value over directly creating a `Notification`?

amc94 (changes requested):
one quick change

backend/test_observer/common/review_notification.py
almeidaraul: you're right, this is a symptom from the refactor. thanks

backend/test_observer/controllers/test_executions/start_test.py
```diff
@@ -121,19 +135,30 @@ def _assign_reviewers_to_environments(self) -> None:
...
+            reviewers_to_assignment_count[reviewer.id] += 1
 
-        self.db.commit()
+        with self.db.begin_nested():
+            batch_create_review_notifications(
+                self.db,
+                newly_assigned_environment_reviewers,
+                self.artefact,
+                NotificationType.USER_ASSIGNED_ENVIRONMENT_REVIEW,
+            )
```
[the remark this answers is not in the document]
almeidaraul: so that's why I had done the dict instead of the list.....

backend/test_observer/common/review_notification.py
```diff
@@ -0,0 +1,113 @@
...
+    for reviewer in reviewers:
+        notification = _create_notification_for_reviewer(
+            user_id=reviewer.id, notification_type=notification_type, target_url=get_artefact_url(artefact)
+        )
+        notifications.append(notification)
+    db.add_all(notifications)
+
+
+def batch_create_jira_reviewer_cards(
+    new_reviewers: BatchReviewerAssignedMessage,
```
wctaylor: [nitpick]
```suggestion
    message: BatchReviewerAssignedMessage,
```

backend/test_observer/controllers/test_executions/start_test.py
```diff
@@ -121,19 +135,30 @@ def _assign_reviewers_to_environments(self) -> None:
...
             if env_review.reviewers and env_review.reviewers[0] in self.artefact.reviewers:
                 continue
             if reviewers_to_assignment_count[reviewers_sorted[current_reviewer].id] >= reviews_per_reviewer:
                 current_reviewer += 1
-            env_review.reviewers = [reviewers_sorted[current_reviewer]]
-            reviewers_to_assignment_count[reviewers_sorted[current_reviewer].id] += 1
+            reviewer = reviewers_sorted[current_reviewer]
+            env_review.reviewers = [reviewer]
+            newly_assigned_environment_reviewers.append(reviewer)
+            reviewers_to_assignment_count[reviewer.id] += 1
```
wctaylor: [nitpick]
The variable naming here is getting confusing. `current_reviewer` is an integer, not a `User`, but `newly_assigned_environment_reviewers` is a `list[User]`, which suggests `current_reviewer` should be a `User`.

Could be nice to clean this up a bit, and possibly also add some comments about what the goals of these lines are

backend/tests/controllers/artefacts/test_artefacts.py
wctaylor, line 1: Maybe worth updating some of these tests to handle multiple reviewers, just to make sure it behaves as expected and each gets their own notifications

backend/tests/controllers/test_executions/test_start_test.py
```diff
@@ -542,6 +544,72 @@ def test_sets_initial_test_execution_status(self, execute: Execute, start_reques
...
+                Notification.user_id.in_(assigned_reviewer_ids),
+                Notification.notification_type == NotificationType.USER_ASSIGNED_ENVIRONMENT_REVIEW,
+            )
+            .all()
+        )
+
+        # Should have notifications for environment reviews (at least one per reviewer)
+        assert len(notifications) == len(assigned_reviewer_ids)
+        # Verify notifications are for the right type
+        assert all(n.notification_type == NotificationType.USER_ASSIGNED_ENVIRONMENT_REVIEW for n in notifications)
```
wctaylor, line 583: I would like to also check that the notifications are correctly assigned (so that we don't have some weird case where the first users gets all the notifications or something)

wctaylor (changes requested):
Overall this looks pretty good! Just a little bit more cleanup I think

backend/test_observer/controllers/artefacts/artefacts.py
```diff
@@ -304,9 +317,28 @@ def patch_artefact(
...
+        )
+
     db.commit()
+
+    if len(newly_assigned_reviewers) > 0 and artefact.jira_issue is not None:
+        review_assigned_messages = BatchReviewerAssignedMessage(
+            artefact,
+            [(reviewer, [NotificationType.USER_ASSIGNED_ARTEFACT_REVIEW]) for reviewer in newly_assigned_reviewers],
+        )
+        batch_create_jira_reviewer_cards(review_assigned_messages)
```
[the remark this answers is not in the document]
almeidaraul: this will be part of a future card

wctaylor (approved):
Looks good to me

amc94 (approved):
+1

PR #696, opened 2026-04-07
TO-299: permission checking with ArtefactMatchingRules

almeidaraul:
## Description

As the final part of TO-299, this PR adds permission checking with ArtefactMatchingRules for `change_artefact`. As it's a very unusual way to gate access to endpoints, I tried to make the tests very comprehensive.

This PR originally failed the endpoint permissions check because of the unusual gating strategy. The workaround I found was to create a sort of "empty" gate for FastAPI, but it does not in fact require that the permissions check is carried out from inside the function. Let me know if you have a better suggestion!

Relevant point raised by copilot review:

> Authorization is checked against the current artefact state. If the PATCH includes a stage change, the AMR match (and thus permission) can change after the update, but this code does not re-check permissions against the destination stage. This can allow users with permission on one stage to move an artefact into a stage they shouldn’t be able to modify. Consider checking permissions against the post-patch stage when request.stage is set (or requiring permission for both the current and target stage).

In this case I think we can leave the authorization check to just cover the artefact's state pre-modification.

## Resolved issues

TO-299

## Tests

Many tests!!! I think we can trust copilot to review the code logic inside them but please as a reviewer try to see if you can find uncovered corner cases


backend/test_observer/common/permissions.py
```diff
@@ -42,3 +45,112 @@ def permission_checker(
...
+        return None
+
+    # Query for matching AMR IDs using the centralized matching logic
+    matching_amr_ids = db.execute(match_artefact(artefact)).scalars().all()
+
+    if not matching_amr_ids:
+        raise HTTPException(status_code=403, detail="Insufficient permissions")
+
+    # Load the full AMR objects with their teams and permissions
+    from test_observer.data_access.models import ArtefactMatchingRule
```
wctaylor: Is this imported here to avoid circular dependencies somewhere, or could this import just be moved to the top?

backend/test_observer/common/permissions.py
```diff
@@ -42,3 +45,112 @@ def permission_checker(
...
+
+    Raises:
+        HTTPException(403): If user is not authorized
+    """
+    # Honor IGNORE_PERMISSIONS config (aligned with permission_checker)
+    if required_permission.value in IGNORE_PERMISSIONS:
+        return None
+
+    if app and required_permission in app.permissions:
+        return None
```
wctaylor: [nitpick]
Same as above

```suggestion
    if required_permission.value in IGNORE_PERMISSIONS:
        return

    if app and required_permission in app.permissions:
        return
```

backend/test_observer/common/permissions.py
```diff
@@ -42,3 +45,112 @@ def permission_checker(
...
+    # and if those AMRs grant the required permission
+    user_team_ids = {team.id for team in user.teams}
+
+    for rule in matching_rules:
+        rule_team_ids = {team.id for team in rule.teams}
+
+        user_in_rule_team = user_team_ids & rule_team_ids
+        rule_grants_permission = required_permission in rule.grant_permissions
+        if user_in_rule_team and rule_grants_permission:
+            return None
```
wctaylor: [nitpick]
I interpret `return None` to be somewhat semantically different from `return`, even though functionally they are of course the same. To me, `return None` suggests that the `None` is indeed a value that might get used, while just `return` suggests that there's just nothing to do. But this is highly subjective and can be ignored.

```suggestion
            return
```

backend/test_observer/common/permissions.py
```diff
@@ -42,3 +45,63 @@ def permission_checker(
...
+
+    Raises:
+        HTTPException(403): If user is not authorized
+    """
+    if user is None:
+        raise HTTPException(status_code=403, detail="Insufficient permissions")
+
+    if user.is_admin:
+        return None
+
```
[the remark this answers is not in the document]
wctaylor: This might be worthwhile to add at least a few tests that ensure admins can still change things

wctaylor (commented):
Mostly looks pretty good! Some comments for now

wctaylor (approved):
Things are definitely getting complicated, but the test coverage is very nice, so that should at least provide some confidence for any future changes as well. Good job!

PR #697, opened 2026-04-07
To 299/permission checking for reruns

almeidaraul:
## Description

As the final part of TO-299, this PR adds permission checking with ArtefactMatchingRules for `change_rerun`. As it's a very unusual way to gate access to endpoints, I tried to make the tests very comprehensive.

Additionally, gating that for reruns is a bit different from `change_artefact` because in the rerun endpoints we:
- May have multiple artefacts
- May not know upfront what the artefacts are

## Resolved issues

TO-299

## Tests

Many tests!!! I think we can trust copilot to review the code logic inside them but please as a reviewer try to see if you can find uncovered corner cases

### DIFF
--- a/backend/test_observer/controllers/test_executions/reruns.py
+++ b/backend/test_observer/controllers/test_executions/reruns.py
@@ -188,7 +188,8 @@ def _validate_amr_permissions_for_request(
     # The selectinload patterns in check_amr_permission() are sufficient
     # to handle the relationships it needs.
     affected_artefacts_query = (
-        select(Artefact).distinct()
+        select(Artefact)
+        .distinct()
         .join(ArtefactBuild, Artefact.id == ArtefactBuild.artefact_id)
         .join(TestExecution, TestExecution.artefact_build_id == ArtefactBuild.id)
         .where(or_(*conditions))
--- a/backend/test_observer/controllers/test_executions/reruns.py
+++ b/backend/test_observer/controllers/test_executions/reruns.py
@@ -23,7 +23,11 @@ from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.orm import Session, selectinload
 
 from test_observer.common.enums import Permission
-from test_observer.common.permissions import check_amr_permission, permission_checker
+from test_observer.common.permissions import (
+    check_amr_permission,
+    permission_checker,
+)
+from test_observer.common.config import IGNORE_PERMISSIONS
 from test_observer.controllers.applications.application_injection import (
     get_current_application,
 )
@@ -154,14 +158,15 @@ def _validate_amr_permissions_for_request(
     permission: Permission,
 ) -> None:
     """
-    Validate AMR permissions for rerun requests.
+    Validate AMR permissions for rerun requests with optimized batch querying.
 
     Handles two cases:
     1. Direct IDs: When request.test_execution_ids is provided
     2. Filters: When request.test_results_filters is provided
 
     Uses all-or-nothing semantics: raise HTTPException(403) if ANY artefact is unauthorized.
-    Bypasses AMR checking if app has the permission or user is admin.
+    Honors IGNORE_PERMISSIONS, app.permissions, and admin bypass through early exits.
+    Uses batch querying to reduce database load from O(n) to O(1) for AMR matching.
 
     Args:
         db: Database session
@@ -173,7 +178,11 @@ def _validate_amr_permissions_for_request(
     Raises:
         HTTPException(403): If any affected artefact is unauthorized
     """
-    # If app has permission, skip AMR checking
+    # Early exit: if permission is in IGNORE_PERMISSIONS, skip all checks
+    if permission.value in IGNORE_PERMISSIONS:
+        return
+
+    # Early exit: if app has the permission, skip AMR checking
     if app and permission in app.permissions:
         return
 
@@ -182,14 +191,6 @@ def _validate_amr_permissions_for_request(
 
     # If no conditions, nothing to check
     if not conditions:
-        return
-
-    # If no user, deny
-    if user is None:
-        raise HTTPException(status_code=403, detail="Insufficient permissions")
-
-    # If user is admin, allow
-    if user.is_admin:
         return
 
     # Query for all unique artefacts affected by the request
@@ -203,12 +204,22 @@ def _validate_amr_permissions_for_request(
 
     affected_artefacts = db.scalars(affected_artefacts_query).all()
 
+    # If no affected artefacts, no permissions to check (e.g., invalid test_execution_ids)
     if not affected_artefacts:
+        return
+
+    # Now check user authorization (only if we have artefacts to check)
+    # Early exit: if no user, deny (unless app permission or IGNORE_PERMISSIONS already passed)
+    if user is None:
+        raise HTTPException(status_code=403, detail="Insufficient permissions")
+
+    # Early exit: if user is admin, allow
+    if user.is_admin:
         return
 
     # Batch match all artefacts to their AMRs in a single query
     batch_match_result = db.execute(batch_match_artefacts(affected_artefacts)).all()
-    
+
     # Build map: artefact_id → set of matching AMR IDs
     artefact_to_amrs: dict[int, set[int]] = {}
     matched_amr_ids = set()
@@ -232,15 +243,15 @@ def _validate_amr_permissions_for_request(
 
     # Check permission for each affected artefact using all-or-nothing semantics
     user_team_ids = {team.id for team in user.teams}
-    
+
     for artefact in affected_artefacts:
         # Check if this artefact has any matching AMRs
         matching_amr_ids = artefact_to_amrs.get(artefact.id, set())
-        
+
         if not matching_amr_ids:
             # No AMRs match this artefact
             raise HTTPException(status_code=403, detail="Insufficient permissions")
-        
+
         # Check if user is in a team with a matching AMR that grants the permission
         has_permission = False
         for amr_id in matching_amr_ids:
@@ -250,9 +261,10 @@ def _validate_amr_permissions_for_request(
                 if user_team_ids & rule_team_ids:
                     has_permission = True
                     break
-        
+
         if not has_permission:
             raise HTTPException(status_code=403, detail="Insufficient permissions")
+
 
 
 
--- a/backend/test_observer/controllers/test_executions/reruns.py
+++ b/backend/test_observer/controllers/test_executions/reruns.py
@@ -25,6 +25,7 @@ from sqlalchemy.orm import Session, selectinload
 from test_observer.common.config import IGNORE_PERMISSIONS
 from test_observer.common.enums import Permission
 from test_observer.common.permissions import (
+    openapi_scope_declaration,
     permission_checker,
 )
 from test_observer.controllers.applications.application_injection import (
@@ -261,6 +262,7 @@ def _validate_amr_permissions_for_request(
     "/reruns",
     response_model=list[PendingRerun] | None,
     dependencies=[
+        Security(openapi_scope_declaration, scopes=[Permission.change_rerun]),
         Security(require_bulk_permission, scopes=[Permission.change_rerun_bulk]),
     ],
 )
@@ -380,6 +382,7 @@ def get_rerun_requests(
 @router.delete(
     "/reruns",
     dependencies=[
+        Security(openapi_scope_declaration, scopes=[Permission.change_rerun]),
         Security(require_bulk_permission, scopes=[Permission.change_rerun_bulk]),
     ],
 )
--- a/backend/test_observer/controllers/test_executions/reruns.py
+++ b/backend/test_observer/controllers/test_executions/reruns.py
@@ -34,6 +34,7 @@ from test_observer.data_access.models import (
     Application,
     Artefact,
     ArtefactBuild,
+    ArtefactMatchingRule,
     Environment,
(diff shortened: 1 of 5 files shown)


backend/test_observer/controllers/test_executions/reruns.py
```diff
@@ -124,11 +149,126 @@ def require_bulk_permission(
...
+
+    Raises:
+        HTTPException(403): If any affected artefact is unauthorized
+    """
+    # Early exit: if permission is in IGNORE_PERMISSIONS, skip all checks
+    if permission.value in IGNORE_PERMISSIONS:
+        return
+
+    # Early exit: if app has the permission, skip AMR checking
+    if app and permission in app.permissions:
```
motjuste: nit: `if app is not None and permission in app.permissions`

Python is very dynamic with truthiness.

backend/test_observer/controllers/test_executions/reruns.py
```diff
@@ -124,11 +149,126 @@ def require_bulk_permission(
...
+
+    affected_artefacts = db.scalars(affected_artefacts_query).all()
+
+    # If no affected artefacts, no permissions to check (e.g., invalid test_execution_ids)
+    if not affected_artefacts:
+        return
+
+    # Now check user authorization (only if we have artefacts to check)
+    # Early exit: if no user, deny (unless app permission or IGNORE_PERMISSIONS already passed)
+    if user is None:
```
motjuste: I believe we should check for user being unauthorised before running any queries to the DB at all.

backend/test_observer/controllers/test_executions/reruns.py
```diff
@@ -124,11 +149,126 @@ def require_bulk_permission(
...
+        raise HTTPException(status_code=403, detail="Insufficient permissions")
+
+    # Early exit: if user is admin, allow
+    if user.is_admin:
```
motjuste: why run the DB query if we are going to allow admin users

backend/test_observer/controllers/test_executions/reruns.py
```diff
@@ -124,11 +149,126 @@ def require_bulk_permission(
...
+
+        if not matching_amr_ids:
+            # No AMRs match this artefact
+            raise HTTPException(status_code=403, detail="Insufficient permissions")
+
+        # Check if user is in a team with a matching AMR that grants the permission
+        has_permission = False
+        for amr_id in matching_amr_ids:
+            rule = amr_rules.get(amr_id)
+            if rule and permission in rule.grant_permissions:
```
motjuste: nit: `if rule is not None and permission in ...`

backend/test_observer/data_access/queries.py
```diff
@@ -83,3 +83,42 @@ def match_artefact(artefact: Artefact) -> Select[tuple[int]]:
     )
 
     return select_rules
+
+
+def batch_match_artefacts(artefacts: list[Artefact]) -> Select[tuple[int, int]]:
```
motjuste, line 88: I am not well versed enough with sqlalchemy to understand this ... I hope this works.

motjuste (changes requested):
Requesting some minor looking changes... haven't looked at the tests yet in case there are more changes.

backend/test_observer/controllers/test_executions/reruns.py
```diff
@@ -124,11 +149,118 @@ def require_bulk_permission(
...
+
+
 @router.post(
     "/reruns",
     response_model=list[PendingRerun] | None,
     dependencies=[
-        Security(permission_checker, scopes=[Permission.change_rerun]),
         Security(require_bulk_permission, scopes=[Permission.change_rerun_bulk]),
     ],
 )
```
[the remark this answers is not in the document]
motjuste: @almeidaraul ... please comment on this and help me understand why this is not relevant ... and generally whenever dismissing comments in general please.

backend/test_observer/data_access/queries.py
```diff
@@ -83,3 +83,42 @@ def match_artefact(artefact: Artefact) -> Select[tuple[int]]:
...
+                or_(ArtefactMatchingRule.stage == artefact.stage, ArtefactMatchingRule.stage == ""),
+                or_(ArtefactMatchingRule.track == artefact.track, ArtefactMatchingRule.track == ""),
+                or_(ArtefactMatchingRule.branch == artefact.branch, ArtefactMatchingRule.branch == ""),
+                or_(ArtefactMatchingRule.name == artefact.name, ArtefactMatchingRule.name == ""),
+            )
+        )
+        queries.append(query)
+
+    # UNION all queries into one
+    return select(union_all(*queries).subquery())
```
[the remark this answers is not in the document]
motjuste: I think this comment is worth addressing with some justification ... esp. considering we've been having DB perf issues at TO.

I mean ... do you ensure that the batch size is small?

backend/tests/conftest.py
```diff
@@ -175,8 +175,12 @@ def override_permissions(*permissions: Permission):
 
 
 def make_authenticated_request(request_func: Callable[[], Response], *permissions: Permission):
-    # First, make sure the endpoint returns 403 without permissions
-    assert request_func().status_code == 403
+    # Verify the endpoint denies unauthenticated access (usually 403)
+    # but allow other error codes (e.g., 404 for missing data, 422 for validation)
```
motjuste: NOPE!

I am sorry but [403](https://http.cat/403) is a completely different animal than [404](https://http.cat/404) and absolutely not [422](https://http.cat/422).

Unauthorized requests mean only one thing ... the caller tried to do something they are not allowed to do ... so they must always be `Forbidden`, nothing else.

backend/test_observer/controllers/test_executions/reruns.py
```diff
@@ -124,11 +149,118 @@ def require_bulk_permission(
...
+        return
+
+    # Collect conditions to find affected TestExecutions
+    conditions = _build_rerun_conditions(request)
+
+    # If no conditions, nothing to check
+    if not conditions:
+        return
+
+    # Check user authorization
```
motjuste, line 198: fmt: misplaced comment

backend/tests/controllers/test_executions/test_reruns.py
```diff
@@ -182,14 +186,14 @@ def test_execution_to_pending_rerun(test_execution: TestExecution) -> dict:
...
+def test_post_no_data_returns_422(test_client: TestClient):
+    response = test_client.post(reruns_url, json=None)
+    assert response.status_code == 422
 
 
-def test_post_invalid_id_returns_404_with_message(post: Post):
-    response = post({"test_execution_ids": [1]})
-    assert response.status_code == 404
-    assert response.json()["detail"] == "Didn't find test executions with provided ids"
+def test_post_invalid_id_returns_403(test_client: TestClient):
```
motjuste: I don't know if this is testing if the user is unauthenticated or the id is wrong.  The test says the latter ... but the status code expected is the former.

I recommend having one test for each separately.

backend/tests/controllers/test_executions/test_reruns.py
```diff
@@ -1094,14 +1092,7 @@ def test_delete_with_empty_ids_does_nothing(test_client: TestClient, post: Post,
...
     # Delete with empty list should do nothing
-    response = make_authenticated_request(
-        lambda: test_client.request(
-            "DELETE",
-            reruns_url,
-            json={"test_execution_ids": []},
-        ),
-        Permission.change_rerun,
-    )
+    response = delete({"test_execution_ids": []})
```
motjuste, line 1095: Please check again the use of `delete` ... I don't think it uses change_rerun_bulk permission.  I don't know if we want to base our testing on `change_rerun` (non-bulk) when that is now also getting removed from the OpenAPI.

backend/tests/controllers/test_executions/test_reruns.py
```diff
@@ -1135,3 +1126,525 @@ def test_post_non_silent_with_single_id_and_filters_fails(test_client: TestClien
 
     assert response.status_code == 422
     assert "must be done silently" in response.json()["detail"]
+
+
+class TestRerunAMRPermissions:
```
motjuste, line 1131: There's a lot of repetition in the test cases ... perhaps we could refactor these out to helper methods ... will also highlight actual difference between the tests more clearly.

motjuste, line 1131: BTW ... good test coverage!

motjuste (changes requested):
Unfortunately ... there's one deal-breaker ... but I believe we have covered things well.

Please also respond to comments from Copilot.  You don't need to add comments to the code unless that'd be useful, but the PRs and comments also serve as documentation.

almeidaraul:
@motjuste

> Please also respond to comments from Copilot. You don't need to add comments to the code unless that'd be useful, but the PRs and comments also serve as documentation.

I only resolved the comments because they were already addressed in https://github.com/canonical/test_observer/pull/696 , sorry about the confusion

motjuste:
> @motjuste
> 
> > Please also respond to comments from Copilot. You don't need to add comments to the code unless that'd be useful, but the PRs and comments also serve as documentation.
> 
> I only resolved the comments because they were already addressed in #696 , sorry about the confusion

no problem ... but that's a good context to have already added ... I hadn't looked at the other PRs.

backend/test_observer/controllers/test_executions/reruns.py
almeidaraul: not misplaced, the code below this checks the user authorization

backend/tests/controllers/test_executions/test_reruns.py
almeidaraul: But it does use change_rerun

motjuste: But not the 'bulk' one ... that permission doesn't get tested against. I don't know ... the previous impl was explicit and clearer.

backend/test_observer/controllers/test_executions/reruns.py
```diff
@@ -124,11 +149,118 @@ def require_bulk_permission(
...
+
+    batch_match_result = db.execute(batch_match_artefacts(list(affected_artefacts))).all()
+
+    # Build map: artefact_id → set of matching AMR IDs
+    artefact_to_amrs: dict[int, set[int]] = {}
+    matched_amr_ids = set()
+    for artefact_id, amr_id in batch_match_result:
+        artefact_to_amrs.setdefault(artefact_id, set()).add(amr_id)
+        matched_amr_ids.add(amr_id)
+
```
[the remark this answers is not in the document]
almeidaraul: I don't think there's a use case where this optimization is relevant and it increases code complexity

backend/tests/conftest.py
```diff
@@ -175,8 +175,9 @@ def override_permissions(*permissions: Permission):
 
 
 def make_authenticated_request(request_func: Callable[[], Response], *permissions: Permission):
-    # First, make sure the endpoint returns 403 without permissions
-    assert request_func().status_code == 403
+    # Verify the endpoint denies unauthenticated access
```
motjuste, line 178: nit: we could simply revert the change since original behaviour was correct, but the comment explains things ... so up to you.

backend/tests/controllers/test_executions/test_reruns.py
motjuste: so we're keeping the `delete` function that uses `Permission.change_rerun` instead of `Permissiong.change_rerun_bulk` ...?  maybe there's we add an additional test with an analogous `delete` function using the `...bulk` permission to ensure that the behaviour is expected with those permission too.

Would be nice to have, but not a blocker.

motjuste (dismissed):
Minor questions / nitpicks from me ... up to you if you address them.

**but** please ensure that all the previous comments, e.g. from copilot are replied to / addressed before merging this.

backend/tests/controllers/test_executions/test_reruns.py
motjuste: I was wrong ... didn't read the `delete` function fully ... it does use [`change_rerun_bulk`](https://github.com/canonical/test_observer/blob/a452530ec0c614235bad0bbdb83c9312f781e3e0/backend/tests/controllers/test_executions/test_reruns.py#L101)

backend/tests/conftest.py
motjuste: not necessary though ...

backend/test_observer/data_access/queries.py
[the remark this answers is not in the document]
almeidaraul: we will not have this much artefacts in the query

backend/tests/controllers/test_executions/test_reruns.py
```diff
@@ -182,14 +186,14 @@ def test_execution_to_pending_rerun(test_execution: TestExecution) -> dict:
...
+def test_post_no_data_returns_422(test_client: TestClient):
+    response = test_client.post(reruns_url, json=None)
+    assert response.status_code == 422
 
 
-def test_post_invalid_id_returns_404_with_message(post: Post):
-    response = post({"test_execution_ids": [1]})
+def test_post_invalid_id_returns_404(post: Post):
+    response = post({"test_execution_ids": [99999]})
     assert response.status_code == 404
```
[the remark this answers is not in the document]
almeidaraul: not necessary

motjuste (approved):
I understand that since the last review you only reverted changes for `change_rerun`, the old permission / scope.  Hoping this works ... +1

PR #704, opened 2026-04-14
[IQA-3564] Optionally require auth for /me endpoints

wctaylor:
## Description

This PR is split from https://github.com/canonical/test_observer/pull/674
This PR adds the `/v1/users/me` and `/v1/applications/me` endpoints behind the authentication checking mechanism.

If authentication is not required, these will just return `null` when accessed without authentication and will return the proper data when accessed with authentication.

If authentication is required, they will error when accessed without authentication and will return the proper data when accessed with authentication. Additionally, an authenticated app will error if trying to access `/v1/users/me`, and an authenticated user will error if trying to access `/v1/applications/me`.

## Resolved issues

IQA-3564

## Web service API changes

`/v1/users/me` and `/v1/applications/me` can optionally require authentication, which in principle will allow users and applications to view themselves even without the `view_user` or `view_application` permission. This works for applications, but because the frontend doesn't expose the `/v1/users/me` endpoint, in practice this means authenticated users can't easily view their own info. We might want to consider adding something to the frontend that exposes a user's own data to themselves, similar to how we do it in Weebl.

## Tests

Unit tests have been added, and manual testing suggests things work as well.

It's worth noting that there's not really a good way to manually test `/v1/users/me` in the authenticated scenario, because directly accessing that endpoint in a browser doesn't send the CSRF token.

### DIFF
--- a/backend/test_observer/controllers/applications/applications.py
+++ b/backend/test_observer/controllers/applications/applications.py
@@ -18,7 +18,7 @@ from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from test_observer.common.enums import Permission
-from test_observer.common.permissions import permission_checker
+from test_observer.common.permissions import permission_checker, requires_authentication
 from test_observer.controllers.applications.application_injection import (
     get_current_application,
 )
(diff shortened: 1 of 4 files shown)


rpbritton (dismissed):
Seems very reasonable to me.

rpbritton (approved):

PR #713, opened 2026-04-18
Add miscellaneous permission tests

wctaylor:
## Description

This PR just adds a few additional permissions-related tests that I thought of while doing recent work on authentication- and permissions-related tasks.

## Tests

Yep

### DIFF
--- a/backend/tests/controllers/applications/test_applications.py
+++ b/backend/tests/controllers/applications/test_applications.py
@@ -15,7 +15,9 @@
 
 from collections.abc import Callable
 
+import pytest
 from fastapi.testclient import TestClient
+from sqlalchemy.exc import ProgrammingError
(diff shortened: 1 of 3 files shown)


motjuste (commented):
I am not sure if [422](https://http.cat/422) is inappropriate ... unsure what we want to tell the user ... your payload is wrong ... or [403](https://http.cat/403) is better ... saying you do not have permission with whatever you are sending.

I think it is the former ... but just need confirmation before +1.

motjuste (approved):
Thinking about it ... I imagine the situation is that ... the user is authorised to make the change in all these tests, but the change they try to make it invalid ... so 422.

Please verify my understanding ... giving a +1

wctaylor:
Yes, exactly. They are authorized, but the change is invalid

PR #716, opened 2026-04-20
Fix: Exports the product artifact deep dive dashboard in a compatible format

moisesbenzan:
Updates the exported grafana dashboard to use the legacy format as the COS grafana version is using version 12.0.X.


wctaylor (commented):
I don't particularly like having some internal identifiers (Juju model and app and stuff) being hard-coded into this, especially since other deployments really won't care about this at all.

If we need to do this just to get something initially working, I can accept that, but I'm wondering if we could somehow take a different approach that decouples the metrics of interest from the code itself. That is, Test Observer should offer a way to expose and export metrics, but beyond that, metrics-related stuff happens outside the repo.

Thoughts?

wctaylor (approved):
Thanks for the tweak!

PR #730, opened 2026-04-29
Spread Again

wctaylor:
## Description

This is a follow-up to https://github.com/canonical/test_observer/pull/729, which misidentified the source of the problem. It seems that it's actually a race condition between the redis and database relations, due to how the charm updates the app environment. The charm needs to be updated, but in the meantime, relating to redis first, before waiting for the database, seems to help. I ran the job three times in a row without it failing, so let's hope this works for now.

## Tests

I ran the CI job three times in a row, and it passed all three times. The fail rate was >50% prior to this (maybe closer to 90%), so I think this helps.

### DIFF
--- a/spread/integration/task.yaml
+++ b/spread/integration/task.yaml
@@ -44,7 +44,10 @@ execute: |
     --config test-observer-api-scheme='http://' \
     --resource frontend-image="${frontend_oci_image}"
 
+  # These two waits are temporary until the charm
+  # is updated to handle the relations and app environment variables more gracefully
   juju wait-for application db --timeout=10m
+  juju wait-for application redis --timeout=10m
 
   juju relate backend db
   juju relate backend redis
--- a/spread/integration/task.yaml
+++ b/spread/integration/task.yaml
@@ -18,7 +18,7 @@
 summary: Run the charm integration test
 
 execute: |
-  set -e
+  set -o pipefail
 
   function crashdump() {
     trap - ERR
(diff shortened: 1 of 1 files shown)


spread/integration/task.yaml
```diff
@@ -18,50 +18,43 @@
...
+
+    microk8s config > kube-config
+    pipx run --spec git+https://github.com/canonical/juju-k8s-crashdump.git juju-k8s-crashdump \
+      ./kube-config concierge-microk8s -o juju-k8s-crashdump.tar.gz
+  }
+
+  trap crashdump ERR
+
   backend_oci_image="$(cat "${SPREAD_PATH}"/backend/charm/charmcraft.yaml | yq '.resources.api-image.upstream-source')"
   frontend_oci_image="$(cat "${SPREAD_PATH}"/frontend/charm/charmcraft.yaml | yq '.resources.frontend-image.upstream-source')"
```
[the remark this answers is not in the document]
wctaylor: I'm not concerned

spread/integration/task.yaml
```diff
@@ -18,50 +18,43 @@
...
 
   juju deploy --trust postgresql-k8s --channel=14/stable --revision=774 db
   juju deploy redis-k8s --channel=latest/edge --revision=27 redis
```
LucasPluta, line 37: I understand why we’re using latest/edge instead of a more stable channel, but how can we inspect the release history for this charm?

wctaylor: Redis is actually deprecated and only receiving occasional bug fix updates. This revision matches what we actually use. I want to swap to a ValKey charm in the coming cycle, but in the meantime, you can look at the tags on this repo to see the various revision releases for redis:
https://github.com/canonical/redis-k8s-operator

motjuste (approved):
+1 ... as many as lines removed!

PR #734, opened 2026-05-04
Feat new terraform

amc94:
## Description
Adds a terraform module that deploys the core  of test observer

## Resolved issues
outdated TF modules

## Documentation
Included in the PR

## Tests
Terraform plan against existing deployment


terraform/core/variables.tf
```diff
@@ -0,0 +1,144 @@
+variable "juju_model_uuid" {
+  description = "UUID of the Juju model to deploy to"
+  type        = string
+}
+######### API #########
+variable "api_config" {
+  description = "Full configuration for the test observer front end"
```
wctaylor: This is `api_config` but the description says "front end"

terraform/core/integrations.tf
wctaylor, line 1: I think the relations between the backend/API and otelcol are missing.

```
api:grafana-dashboard otelcol:grafana-dashboards-consumer
api:metrics-endpoint otelcol:metrics-endpoint
```

These would also be optional

wctaylor (commented):
Looks quite good to me! Just a few things

wctaylor (approved):
Some Copilot comments might be worth considering in the future, but for now, I think this is an excellent start, and it looks good to me

PR #738, opened 2026-05-12
Feat: partial environment name matching in test results and test execution endpoints

moisesbenzan:
## Description

Adds a new `environment_contains` query parameter to three backend endpoints, enabling partial, case-insensitive environment name matching scoped by artefact family.

Previously, filtering by environment required an exact name match (`environments=some-env-name`). This was too strict for workflows where users want to filter across environments that share a common substring (e.g. all `desktop-*` environments, or all `*-jammy` environments).

The new filter accepts multiple values that are AND'd together — each value must be a substring of the environment name. Combined with the existing `families` filter, the matching is naturally scoped to the desired artefact family.

Implementation: each value in `environment_contains` is applied as an `ILIKE '%value%'` condition on `Environment.name`. Multiple values produce multiple `WHERE` clauses joined by `AND`.

## Resolved issues
N/A

## Documentation
No public documentation impact. The new query parameter is self-describing and will appear in the auto-generated OpenAPI docs at `/docs`.

## Web service API changes
Three existing endpoints gain a new optional query parameter `environment_contains: list[str]:`

**GET /v1/environments**

- New param: `environment_contains` (repeated query param, e.g. `?environment_contains=desktop&environment_contains=jammy`)
- Filters the returned environment names to those containing all provided substrings (case-insensitive)
- Can be combined with the existing families and q parameters
- No auth changes; existing view_test permission applies

**GET /v1/test-results**

- New param: `environment_contains`
- Filters test results to those whose environment name contains all provided substrings
- ANDed with all other existing filters (e.g. `families`, `artefacts`)

**GET /v1/test-executions**

New param: `environment_contains`
Same semantics as above for test executions
No new endpoints, no DB migrations, no configuration changes. The new filter adds a `JOIN` to the environments table only when `environment_contains` is provided, consistent with how the existing environments filter works.

No performance regression expected — `Environment.name` is already indexed.

## Tests

New tests added in:

- `backend/tests/controllers/environments/test_environments.py` — 5 tests covering single value match, multiple values `AND`'d, case-insensitivity, no-match empty result, and family scoping
- `backend/tests/controllers/test_results/test_test_results.py` — 4 tests in `TestEnvironmentContainsFilter` covering the same scenarios for the test results endpoint
- `backend/tests/controllers/test_executions/test_search.py` — 4 tests in `TestEnvironmentContainsFilter` covering the same scenarios for the test executions search endpoint

To run:

```
cd backend
uv run python -m pytest tests/controllers/environments/test_environments.py tests/controllers/test_results/test_test_results.py::TestEnvironmentContainsFilter tests/controllers/test_executions/test_search.py::TestEnvironmentContainsFilter -v
```


backend/test_observer/common/helpers.py
```diff
@@ -35,3 +35,20 @@ def get_artefact_url(artefact: Artefact) -> str:
...
+    if not values:
+        return []
+
+    terms: list[str] = []
+    for value in values:
+        stripped = value.strip()
+        if not stripped:
+            continue
+
+        escaped = stripped.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
```
almeidaraul: this isn't enough input sanitation, I think we should instead only allow valid characters instead of filtering out some. otherwise we'll have an issue at some point

backend/tests/common/test_helpers.py
```diff
@@ -0,0 +1,29 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+# SPDX-FileCopyrightText: Copyright 2026 Canonical Ltd.
+# SPDX-License-Identifier: AGPL-3.0-only
+
+from test_observer.common.helpers import normalize_contains_terms
```
almeidaraul: then if you agree with my previous comments I would also add some tests here for valid inputs like "a23-322a" or whatever

backend/tests/controllers/test_executions/test_search.py
```diff
@@ -476,3 +476,152 @@ def test_filter_by_event_names_mixed_none_and_specific_returns_422(
...
+            lambda: test_client.get(
+                "/v1/test-executions",
+                params=[("test_result", "none"), ("environment_contains", ""), ("environment_contains", "   ")],
+            ),
+            Permission.view_test,
+        )
+        assert response.status_code == 200
+        assert response.json()["test_executions"] == []
+
+    def test_like_wildcards_are_treated_as_literals(self, test_client: TestClient, generator: DataGenerator):
```
almeidaraul: maybe the change I suggested affects this test

backend/tests/controllers/test_results/test_test_results.py
```diff
@@ -1673,3 +1673,169 @@ def test_execution_is_latest_different_artefact_builds(self, test_client: TestCl
...
+            lambda: test_client.get(
+                "/v1/test-results",
+                params=[("environment_contains", ""), ("environment_contains", "   ")],
+            ),
+            Permission.view_test,
+        )
+        assert response.status_code == 200
+        assert response.json()["test_results"] == []
+
+    def test_like_wildcards_are_treated_as_literals(self, test_client: TestClient, generator: DataGenerator):
```
almeidaraul: reminder to check this test

backend/tests/controllers/environments/test_environments.py
```diff
@@ -332,3 +333,177 @@ def test_get_environments_pagination_metadata(test_client: TestClient, generator
...
+            "/v1/environments",
+            params=[("environment_contains", ""), ("environment_contains", "   ")],
+        ),
+        Permission.view_test,
+    )
+    assert resp.status_code == 200
+    assert resp.json()["environments"] == []
+
+
+def test_environment_contains_treats_like_wildcards_as_literals(test_client: TestClient, generator: DataGenerator):
```
almeidaraul: please check this as well

almeidaraul (changes requested):

almeidaraul (changes requested):
I'm ready to approve but it seems CI is failing, can you check that?

backend/test_observer/controllers/test_results/filter_test_results.py
```diff
@@ -80,6 +80,14 @@ def build_query_filters_and_joins(
         query_filters.append(Environment.name.in_(filters.environments))
         joins_needed.update(["test_execution", "environment"])
 
+    if len(filters.environment_contains) > 0:
+        for value in filters.environment_contains:
+            normalized_value = value.strip()
+            if not normalized_value:
+                continue
+            query_filters.append(Environment.name.icontains(normalized_value, escape="\\"))
```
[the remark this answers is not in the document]
moisesbenzan: autoescape is true by default

backend/test_observer/controllers/test_executions/search.py
```diff
@@ -186,6 +186,14 @@ def _build_execution_filters(
         query_filters.append(Environment.name.in_(filters.environments))
         joins_needed.add("environment")
 
+    if filters.environment_contains:
+        for value in filters.environment_contains:
+            normalized_value = value.strip()
+            if not normalized_value:
+                continue
+            query_filters.append(Environment.name.icontains(normalized_value, escape="\\"))
```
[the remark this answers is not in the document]
moisesbenzan: autoescape is true by default

backend/test_observer/controllers/environments/environments.py
```diff
@@ -86,6 +90,14 @@ def get_environments(
         search_term = f"%{q.strip()}%"
         query = query.where(Environment.name.ilike(search_term))
 
+    # Apply environment_contains filter (AND'd ICONTAINS per value)
+    if environment_contains:
+        for value in environment_contains:
+            normalized_value = value.strip()
+            if not normalized_value:
+                continue
+            query = query.where(Environment.name.icontains(normalized_value, escape="\\"))
```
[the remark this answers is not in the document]
moisesbenzan: autoescape is true by default

backend/test_observer/controllers/test_results/test_results.py
```diff
@@ -183,6 +183,12 @@ def search_test_results(
         list[str] | None,
         Query(description="Filter by environment names"),
     ] = None,
+    environment_contains: Annotated[
```
rpbritton, line 186: Is this necessary for searching test results? I feel like this opens us up to adding a LOT of filters. Considering that there are not a crazy amount of environments, is it reasonable just to search environments with `GET /environments` and pass in the hard coded values here?

I'm also worried there could be some performance impact since this is already a heavy endpoint, but maybe postgresql is good at that sort of thing.

Would be interested in a metric.

backend/test_observer/controllers/environments/environments.py
```diff
@@ -86,6 +90,14 @@ def get_environments(
         search_term = f"%{q.strip()}%"
```
rpbritton, line 90: Can we just use `icontains` with the existing `q` parameter? I think that will actually make `q` more robust and should be backwards compatible.

rpbritton (changes requested):

backend/test_observer/controllers/test_results/test_results.py
moisesbenzan: fair enough. I'll remove it from this endpoint

moisesbenzan:
Closing this PR as this feature has been implemented in a different endpoint.

PR #742, opened 2026-05-14
Two candidate agentic rewrites of the UI

canonical-vultaire:
This branch contains multiple candidate UIs, intended to be served alongside the current Flutter UI.  The hope is one of these UIs will eventually replace the Flutter UI, but I want to make these available to experiment with side-by-side with the current UI.

They do not interfere with the existing Flutter UI.  There are minimal changes to allow for these UIs to be built and packed along with the Flutter UI.  I would suggest that code review be focused primarily on the modified files for now - while code review for the others is welcome (and a cursory check for "bad behavior" *should* be done!), I would suggest deeper review wait until we're ready to consider a cutover and adoption of the new UI in question, as it is still uncertain if we cut over to one of these implementations or not.


canonical-vultaire:
Closing.  Reviewed internally; we've selected a POC to use as a base and/or reference.

PR #749, opened 2026-05-19
Bump min FastAPI version to 0.120.0 to have starlette 0.49.1 [SECURITY]

motjuste:
## Description

Upgrading vulnerable transitive dependency `starlette` requires updating FastAPI.  Dependabot couldn't do it with our restrictions on compatible FastAPI versions. 

Hence I move the FastAPI version requirement to the minimum that supports the appropriate `starlette` version.  I chose not to go to latest to avoid unintentionally breaking something.

No documented breaking changes impact us when it comes to FastAPI.  The main highlights were FastAPI cloud (which we already avoid), and Pydantic v2 (which we already use).

## Resolved issues

https://github.com/canonical/test_observer/security/dependabot/65

## Documentation

No changes to documentation.

## Web service API changes

None implemented, I hope FastAPI didn't break anything.

## Tests

Shall we see.

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -21,7 +21,7 @@ dynamic = ["version"] # uv-dynamic-versioning used
 requires-python = ">=3.12"
 
 dependencies = [
-    "fastapi>=0.115.0,<0.116.0",
+    "fastapi>=0.120.0,<0.121.0",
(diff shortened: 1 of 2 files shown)


wctaylor (approved):
Looks good to me!

PR #752, opened 2026-05-19
[TO-323] Update setuptools

wctaylor:
## Description

This PR updates the `setuptools` package to address a security issue. This package is ultimately pulled in through `launchpadlib` and its associated dependencies. These dependencies do not specify any `setuptools`, so even updating them doesn't change anything. I generated the lockfile update by using `uv remove launchpadlib` followed by `uv add launchpadlib` and taking the newly generated `setuptools` entry.

## Resolved issues

TO-233


motjuste:
@wctaylor ... Can you try running `uv lock --upgrade-package setuptools` ... you can even specify the version ... I don't believe it produces the exact same result but it seems like it _may_ be correct-er, and it is a single command that you can blame instead of yourself if there's a mistake.

motjuste:
> @wctaylor ... Can you try running `uv lock --upgrade-package setuptools` ... you can even specify the version ... I don't believe it will produce the same result, but in a single command that you can blame instead of yourself if there's a mistake.

BTW ... copied this from with modification from [what dependabot tries to run](https://github.com/canonical/test_observer/actions/runs/26097481817/job/76739020033#step:3:122) ... so I don't get the blame either please.

wctaylor:
@motjuste 
The content of the `setuptools` package itself was unaffected by the command - it did generate the same result. But that command deleted a few other lines elsewhere. And to double check, I first switched to `main` and ran the command from main to see what it would do when updating from the version in `main`. Same result

motjuste (approved):

PR #777, opened 2026-05-22
[IQA-3628] Remove broken workflow

wctaylor:
## Description

This PR removes the scheduled workflow that tries to update charmlibs. This workflow has been running every four hours and failing every time for who knows how long. Since we already have a separate workflow that detects when the charmlibs are out-of-date and tags PRs, I don't think this one is necessary.

## Resolved issues

IQA-3628


amc94 (approved):
+1

PR #786, opened 2026-05-26
Remove redundant risk from artefact

almeidaraul:
## Description

In a previous PR I added the `risk` field to `Artefact` to be used in artefacts of the `solution` family. I now realize we could have just used `stage` to store that information.

This PR removes the `risk` fields and populates unfilled `stage` values with the old `risk` values in the migration. `risk` is also removed from tests and the API

### DIFF
--- a/backend/migrations/versions/2026_05_26_1058-236cc88a9a64_remove_risk_from_artefact.py
+++ b/backend/migrations/versions/2026_05_26_1058-236cc88a9a64_remove_risk_from_artefact.py
@@ -1,4 +1,3 @@
-
 """Remove risk from Artefact
 
 Revision ID: 236cc88a9a64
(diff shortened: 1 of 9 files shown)


backend/migrations/versions/2026_05_26_1058-236cc88a9a64_remove_risk_from_artefact.py
```diff
@@ -0,0 +1,53 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.drop_index("unique_solution", table_name="artefact", postgresql_where="(family = 'solution'::familyname)")
+    op.create_index(
+        "unique_solution",
+        "artefact",
+        ["name", "source", "version", "track", "bundled_builds_hash"],
```
wctaylor: Should `stage` be part of the unique identifier if `risk` was?

almeidaraul: thanks, updated

wctaylor (approved):
Looks good. Kind of a shame that `stage` was there first, as I feel like `risk` is the better term, but oh well.

Also, for posterity, the original PR mentioned in the description is https://github.com/canonical/test_observer/pull/724

PR #791, opened 2026-06-03
Address Starlette BadHost CVE

wctaylor:
## Description

This PR pins `starlette>=1.0.1`, which is the minimum Starlette version required to address [BadHost / CVE-2026-48710](https://badhost.org/). Updating FastAPI to 0.133.0 is required to accommodate Starlette >= 1.0. This PR then fixes a few things that broke as a result of the updates and tweaks some handling of `request.url.path` just to be extra safe.

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
     "python-multipart>=0.0.20,<0.1.0",
     "itsdangerous>=2.2.0",
     "prometheus-fastapi-instrumentator>=7.1.0",
-    "starlette==1.0.1",
+    "starlette>=1.0.1",
 ]
 
 [dependency-groups]
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -21,7 +21,7 @@ dynamic = ["version"] # uv-dynamic-versioning used
 requires-python = ">=3.12"
 
 dependencies = [
-    "fastapi>=0.120.0,<0.121.0",
+    "fastapi>=0.133.0,<0.134.0",
     "uvicorn[standard]>=0.47.0,<0.48.0",
     "sqlalchemy>=2.0.16,<3.0.0",
     "pg8000>=1.29.4,<2.0.0",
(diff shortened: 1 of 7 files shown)


backend/test_observer/users/user_injection.py
```diff
@@ -59,7 +59,9 @@ def get_user_session_browser_friendly(request: Request, db: Session = Depends(ge
     """
 
     # Enforce that this dependency is only used for the /docs endpoint
-    if request.url.path != "/docs":
+    # request.url reconstructs the full URL using the Host header; avoid relying on it for security checks.
```
almeidaraul: maybe start this comment and the other one with "we use `request.scope["path"]` instead of `request.url` because [...]"

almeidaraul (changes requested):

backend/pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     "python-multipart>=0.0.20,<0.1.0",
     "itsdangerous>=2.2.0",
     "prometheus-fastapi-instrumentator>=7.1.0",
+    "starlette==1.0.1",
```
motjuste: Why adding this dependency?  And why so strictly pinned unlike others?

wctaylor: Starlette is already a dependency of FastAPI, and Starlette is the source of the CVE. FastAPI will not automatically pull the required version, so it needs to be manually added to ensure the correct version. I could relax the strictness if desired, though

motjuste: I was hoping that we just bump FastAPI then update lock file ... which _should_ have updated starlette... and if it didn't, we run another update targeting starlette

motjuste: Well, with more context about the CVE ... I now believe we shouldn't  rely on FastAPI to transitively update starlette

backend/test_observer/users/user_injection.py
wctaylor: Addressed

almeidaraul (approved):

PR #796, opened 2026-06-08
[TO-381] Fix scroll jump from text

wctaylor:
## Description

This PR fixes an issue where trying to select text from the IO Log of a test result would cause the scroll bar to jump to the very bottom of the page, effectively making it impossible to actually select text. The cause of this issue seems to be the use of `SelectableText` while already within a `SelectableArea`. Changing it to `Text` (which is consistent with how the `Category` and `Comment` areas are implemented) resolved the problem. 

Once that was fixed, it exposed a different issue, which was that the text selection would overflow and mix in the filter titles from the filter side bar if it was expanded. As a result, I split the filter sidebar and the main body into two separate `SelectionArea`s. 

Finally, I changed the expansion behavior slightly, as scrolling down past an expanded entry would cause it to collapse, resulting in a jarring scrolling effect as the page size instantaneously changed, and which is not intuitive behavior in my opinion. However, that is a separate commit which can easily be removed from this PR if desired.

## Resolved issues

TO-381

## Testing

Manual testing confirms that text selection behaves as expected, and the page no longer jumps to the bottom when clicking within the IO Log.

### DIFF
--- a/frontend/lib/ui/artefact_page/artefact_page.dart
+++ b/frontend/lib/ui/artefact_page/artefact_page.dart
@@ -73,7 +73,9 @@ class ArtefactPage extends ConsumerWidget {
                     Visibility(
                       visible: showSide,
                       maintainState: true,
-                      child: ArtefactPageSide(artefact: artefact),
+                      child: SelectionContainer.disabled(
+                        child: ArtefactPageSide(artefact: artefact),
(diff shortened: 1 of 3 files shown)


almeidaraul (approved):
no need to split into two PRs

PR #810, opened 2026-06-24
TO-388: team permissions should be enough when there are no AMRs with the required permissions

almeidaraul:
## Description

There was a bug where a user belonging to a team with the required permission to access some artefact would not be able to access that if there were no AMRs giving them the permission. This PR fixes that with a simple check for that case.

## Resolved issues

Resolves TO-388

## Tests

Added a regression test and one test for the reverse case.


wctaylor:
Superseded by https://github.com/canonical/test_observer/pull/811

PR #815, opened 2026-06-29
[SQT-200] Add priority column to reruns

rpbritton:
## Description

Adds a priority column to reruns. This is used in the ordering of how reruns are returns.

Also adds the ability to set this to the idempotent rerun creation api, for setting the priority.

Default priority is `0`, with higher priorities being, well, higher priority (returned first).

## Resolved issues

No way to prioritize reruns.

https://warthogs.atlassian.net/browse/TO-200

## Documentation

OpenAPI schema reflects the changes.

## Web service API changes

OpenAPI schema reflects the changes (backwards compatible).

## Tests

Unit tests added for priority cases. Existing Rerun API unit tests pass as well.

### DIFF
--- a/backend/tests/controllers/test_executions/test_reruns.py
+++ b/backend/tests/controllers/test_executions/test_reruns.py
@@ -1695,13 +1695,13 @@ def test_post_without_priority_does_not_change_existing_priority(post: Post, get
     assert reruns[0]["priority"] == 7
 
 
-def test_bulk_post_without_priority_does_not_change_existing_priority(
+def test_silent_post_without_priority_does_not_change_existing_priority(
     test_client: TestClient, get: Get, generator: DataGenerator
 ):
-    """Bulk silent post without priority should not overwrite an existing rerun's priority."""
+    """Silent post without priority should not overwrite an existing rerun's priority."""
     a = generator.gen_artefact(StageName.beta)
     ab = generator.gen_artefact_build(a)
-    e = generator.gen_environment("bulk-no-priority-env")
+    e = generator.gen_environment("silent-no-priority-env")
     te = generator.gen_test_execution(ab, e)
 
     make_authenticated_request(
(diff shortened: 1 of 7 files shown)


backend/tests/controllers/test_executions/test_reruns.py
```diff
@@ -1618,3 +1619,145 @@ def test_no_matching_test_executions_in_filter(self, test_client: TestClient, ge
...
+        Permission.change_rerun,
+    )
+    make_authenticated_request(
+        lambda: test_client.post(reruns_url, params={"silent": True}, json={"test_execution_ids": [te.id]}),
+        Permission.change_rerun,
+    )
+
+    reruns = get().json()
+    assert len(reruns) == 1
+    assert reruns[0]["priority"] == 3
```
wctaylor: This doesn't seem to directly test bulk. The requests should pass multiple test execution IDs, no?

backend/tests/controllers/test_executions/test_reruns.py
```diff
@@ -1618,3 +1619,145 @@ def test_no_matching_test_executions_in_filter(self, test_client: TestClient, ge
...
+            reruns_url,
+            params={"silent": True},
+            json={"test_execution_ids": [te.id], "priority": 9},
+        ),
+        Permission.change_rerun,
+    )
+
+    reruns = get().json()
+    assert len(reruns) == 1
+    assert reruns[0]["priority"] == 9
```
wctaylor: Also seems like this doesn't directly test bulk, unless I'm mistaken

wctaylor (commented):
Mostly looks good! I think two tests aren't quite right, though. Am I wrong?

backend/tests/controllers/test_executions/test_reruns.py
rpbritton: So it's test of `"silent": true` which is the "bulk path" (because returning all the reruns for bulk operations is heavy).

Update the test names to be more clear anyway.

backend/tests/controllers/test_executions/test_reruns.py
rpbritton: Same as above.

wctaylor (approved):

PR #816, opened 2026-06-30
Add created timestamp to rerun response

rpbritton:
## Description

Currently for the test scheduler we use the test execution created timestamp, which may be different from the rerun creation timestamp. This may be wrong.

## Resolved issues

Test execution created timestamp != rerun created timestamp.

## Documentation

N/A

## Web service API changes

Schema up to date.

## Tests

Unit test added.


wctaylor (approved):
Looks sensible to me

PR #846, opened 2026-08-12
fix: priority clarity

### LINKED ISSUE #845: Rerun priority is confusing
Rerun priority should be explained in the docs and the UI text should be more informative

almeidaraul:
## Description

Priority is better explained with docs + a textual hint:

image

## Resolved issues

Fixes #845

### DIFF
--- a/docs/explanation/glossary.rst
+++ b/docs/explanation/glossary.rst
@@ -45,7 +45,7 @@ Includes the name and status (PASSED, FAILED, or SKIPPED) of a test. Additionall
 Rerun request
 -------------
 
-A request to run a test group (a test plan on an artefact build under a particular environment) again. Test Observer does not run reruns itself; it stores the request in a queue that an external test scheduler polls to run the tests and submit new results. See :doc:`test-execution-lifecycle`.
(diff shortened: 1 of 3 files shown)


docs/explanation/glossary.rst
```diff
@@ -41,3 +41,13 @@ Test Result
 -----------
 
 Includes the name and status (PASSED, FAILED, or SKIPPED) of a test. Additionally, it optionally can have logs and other useful bits of information.
+
+Rerun request
+-------------
+
+A request to run a test group (a test plan on an artefact build under a particular environment) again. Test Observer does not run reruns itself; it stores the request in a queue that an external test scheduler polls to run the tests and submit new results. See :doc:`test-execution-lifecycle`.
```
amc94: Test Group isn't defined anywhere

amc94: well it's defined in the parenthesis but that's a confusing definition.

amc94: what's the difference between test group and test execution?

amc94 (changes requested):
why new words when old words work?

docs/explanation/glossary.rst
almeidaraul: a test group does not exist and it's supposed to exist prior to a test execution. I removed this name from the docs now

amc94 (approved):
+1

PR #850, opened 2026-08-18
feat: read solution data from attributes, stop writing legacy fields [3/3]

almeidaraul:
Part 3 of 3 (stacked). **Base: `TO-404/2-write-both` (#849).**

## What
**Migrate + read-new step**: backfill existing data, switch reads to `attributes`, and stop writing the legacy solution-specific fields.

- Backfill migration copies existing `bundled_builds_hash` / bundled build associations into `attributes` for rows written before the write-both code.
- Controllers now read solution identity from `attributes` and stop writing the legacy fields. `StartSolutionTestExecutionRequest` accepts `attributes` (with `track`/`source` kept as deprecated legacy inputs folded into it); `ArtefactPatch` accepts `attributes` and drops `bundled_builds`. `get_artefact_versions` matches on `family` instead of the bundled-builds hash.
- Remove the `bundled_builds` relationship/hash, association table and related ORM event hooks from the model, dropping them from API responses, repository and data generator.

## Safety
The legacy DB column and table are **left in place** (dropped later by a separate destructive/contract migration); `alembic check` ignores them via transitional `include_object` exclusions. Safe to deploy **after** the write-both release during a rolling upgrade.

## Not included (future PR)
The destructive contract migration (drop `bundled_builds_hash` + `artefact_bundled_builds_association`, swap `unique_solution` to `(name, version)`) is a **separate 4th PR** deployed after this one. It stacks on this branch with only a `down_revision` bump to `d315c1f212b9`.

## Review note
Review only against the base branch (#849). Merge/deploy **after** #849.

### DIFF
--- a/.github/workflows/test_backend.yml
+++ b/.github/workflows/test_backend.yml
@@ -57,6 +57,10 @@ jobs:
       - name: Check if alembic migrations are up to date
         run: docker compose exec test-observer-api alembic check
       - name: Compare schema with repository
-        run: diff <(curl -s http://localhost:30000/openapi.json | jq) schemata/openapi.json
+        # Normalise integer-valued floats so the comparison is stable across jq
+        # versions (jq 1.6 on jammy canonicalises 1000000.0 -> 1000000 while jq
+        # 1.7 preserves the literal). Keep this filter in sync with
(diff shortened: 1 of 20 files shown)


almeidaraul:
closing in favor of #876

PR #874, opened 2026-08-28
[New] Missing artefact environments notification backend API

douglasdotc:
## Description

This PR adds the backend API for retrieving missing environments for artefacts.
This API gets the ground truth of the environments required of an artefact from C3 (https://github.com/canonical/hexr/pull/1333) and compares the environment list with the artefact ones, returning the missing environments.

> [!NOTE]
> I referenced https://certification.canonical.com/docs/help/how-to/authenticate/, and it looks like we need a Bearer token for this, and I am not quite sure what the best way is to add the token to this repo.

## Resolved issues

https://warthogs.atlassian.net/browse/CER-3551

## Web service API changes

* Adds C3 to external APIs
* New API `GET /v1/artefacts/{artefact_id}/missing-environments`
Example return:
```
{
  "missing_environments": [
    "cm3p",
    "rpi2",
    "rpi3ap",
    "rpi3b",
    "rpi3bp",
    "rpi4b1g",
    "rpi4b2g",
    "rpi4b4g",
    "rpi4b8g",
    "rpicm4l",
    "rpiz2"
  ],
  "previous_artefact_link": "http://localhost:30001/debs/12?Environment=cm3p&Environment=rpi2&Environment=rpi3ap&Environment=rpi3b&Environment=rpi3bp&Environment=rpi4b1g&Environment=rpi4b2g&Environment=rpi4b4g&Environment=rpi4b8g&Environment=rpicm4l&Environment=rpiz2"
}
```

## Tests

Tests added

### DIFF
--- a/backend/test_observer/common/config.py
+++ b/backend/test_observer/common/config.py
@@ -28,10 +28,6 @@ ADDITIONAL_CORS_ORIGINS = [
     origin.strip() for origin in os.getenv("ADDITIONAL_CORS_ORIGINS", "").split(",") if origin.strip()
 ]
 FRONTEND_URL = os.getenv("FRONTEND_URL", "http://localhost:30001")
-# C3 (certification.canonical.com) is the source of truth for the set of
-# environments (Testflinger queues) an artefact is expected to be tested on.
-# Only used by the missing-environments feature. When C3_API_TOKEN is empty the
-# feature degrades gracefully (source of truth reported as unavailable).
 C3_API_BASE_URL = os.getenv("C3_API_BASE_URL", "https://certification.canonical.com")
 C3_API_TOKEN = os.getenv("C3_API_TOKEN", "")
 SESSIONS_SECRET = os.getenv("SESSIONS_SECRET", "secret")
--- a/backend/test_observer/common/config.py
+++ b/backend/test_observer/common/config.py
@@ -28,6 +28,12 @@ ADDITIONAL_CORS_ORIGINS = [
     origin.strip() for origin in os.getenv("ADDITIONAL_CORS_ORIGINS", "").split(",") if origin.strip()
 ]
 FRONTEND_URL = os.getenv("FRONTEND_URL", "http://localhost:30001")
+# C3 (certification.canonical.com) is the source of truth for the set of
+# environments (Testflinger queues) an artefact is expected to be tested on.
+# Only used by the missing-environments feature. When C3_API_TOKEN is empty the
+# feature degrades gracefully (source of truth reported as unavailable).
+C3_API_BASE_URL = os.getenv("C3_API_BASE_URL", "https://certification.canonical.com")
+C3_API_TOKEN = os.getenv("C3_API_TOKEN", "")
 SESSIONS_SECRET = os.getenv("SESSIONS_SECRET", "secret")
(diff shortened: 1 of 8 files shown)


wctaylor (commented):
How urgent is this functionality? We have some upcoming changes planned for the `Artefact` data model that will add a more generic `attributes` JSONB field. Rather than adding more coupling to C3 and special handling for debs and snaps, I think we could use an approach where the Hardware Certification team instead pushes a `required_environments` (or something similar) key to the `attributes` field when creating/updating artefacts, and then whatever consumer you had in mind for this `missing-environments` endpoint could instead check the current environments against the `required_environments` key

Hook25:
Hey Will, I believe the objective of this is to update the red scary banner with an actual pre-filtered backlink (this artifacts is missing xyz environments. See: link). Please feel free to provide feedback on the idea or the data model, we can ideed change the flow of how the data gets to TO, but I believe (reading it) this felt natural to implement as it is.

wctaylor:
Yeah, it's mostly just not very aligned with the direction we want to take Test Observer. We want Test Observer to become very Artefact-agnostic and not need to know anything about a specific Artefact or change behavior based on an Artefact.

If I look at the original Jira ticket this references, the method proposed there seems quite reasonable to me, but this PR seems to be something different. As I interpret the Jira ticket (and as I think the better solution would be), you could have a job that gets the environments associated with an Artefact in Test Observer (using existing endpoints), and then that job could do the comparison with C3 and send you an alert through some other mechanism.

Hook25:
Now, I didn't read the PR yet, but the idea was to try to contribute this upstream, as in, we communicate, along with the artifacts, the testing expectations (expected environemnts). While those are not met a notification is shown (like the banner currently is) on the card. Additionally, the banner has to be clickable so one doesn't have to have an offline script like I do that compares artifacts. 

If you don't think this is fit for test observer let us know, we will find a way to make it work on our own infra, but it seems like a very big lost opportunity, after all when you are testing something, if you have a list of environments (which is like a coverage criterion) it is useful to be able to define and enforce your coverage criterion.

backend/test_observer/common/config.py
```diff
@@ -28,6 +28,8 @@
     origin.strip() for origin in os.getenv("ADDITIONAL_CORS_ORIGINS", "").split(",") if origin.strip()
 ]
 FRONTEND_URL = os.getenv("FRONTEND_URL", "http://localhost:30001")
+C3_API_BASE_URL = os.getenv("C3_API_BASE_URL", "https://certification.canonical.com")
+C3_API_TOKEN = os.getenv("C3_API_TOKEN", "")
```
gntzio, line 32: The token has a limited lifetime.
Consider authenticating with resolving the token each time before calling C3 API.
See the example: https://github.com/canonical/hwcert-jenkins-jobs/blob/44cdd4a1d006084c33c6f588b643b40991622c46/tools/pools/c3_client.py#L28

Hook25:
As agreed in the conversation in meets, @wctaylor is taking over this task, which will result in either: landing the PR as is or, if we have enough bandwidth, give us a "proper" API that doesn't rely on querying C3 that we can call from our jenkins to provide TO with the testing expectation.

We agreed this makes sense because testing expectations are a key missing feature of the website probably many other would love to have (for example other families) so this eitherr needs a new v2 api that deprecates this one and is more generic or this be implemented "properly" the first time around.

PR #886, opened 2026-09-14
Add security schemes to OpenAPI spec

astrojuanlu:
## Description

Declare the API security schemes in the OpenAPI spec so tooling (e.g. specli) can attach credentials, and document what each endpoint actually accepts:

- `bearerAuth` (application API key) and `sessionCookieAuth` + `csrfTokenAuth` (user session) as security schemes, with a root-level requirement accepting either.
- Per-operation requirements derived from the routes' dependencies: public endpoints (SAML flows, health probes) require nothing; user-only and application-only endpoints advertise only the credential type they accept.
- The spec description notes that authentication is only enforced with `REQUIRE_AUTHENTICATION` enabled (off by default in development).

This also resolves the 91 `security-defined` lint errors in the redocly recommended ruleset.

## Tests

Verified against a local server running with `REQUIRE_AUTHENTICATION=true`.

`specli` with the API key:

```
$ npx specli exec ./backend/schemata/openapi.json --auth bearerAuth --bearer-token "$TO_TOKEN" --server https://test-observer-api.canonical.com artefacts get-v1-artefact-id-get 406435
{
  "id": 406435,
  "name": "mysql-k8s",
  ...
}
```

Without credentials it still fails as expected:

```
$ npx specli exec ./backend/schemata/openapi.json --server https://test-observer-api.canonical.com artefacts get-v1-artefact-id-get 406435
{
  "detail": "Insufficient permissions"
}
```

Note `--auth bearerAuth` is now required to select the scheme, since the spec declares several.

`redocly lint` (was 92 errors / 19 warnings, now 1 pre-existing error / 19 warnings):

```
$ npx --yes @redocly/cli@latest lint ./backend/schemata/openapi.json --format summary
error   no-empty-servers: 1
warning operation-4xx-response: 15
warning no-illogical-composition-keywords: 4

❌ Validation failed with 1 error and 19 warnings.
```


astrojuanlu:
I wish I could request a Copilot review myself and iterate faster, but I can't 😅 thanks for the patience, will continue iterating.

wctaylor:
Thank you for your patience as well! I really like the direction this PR is going, so I appreciate your contribution!

backend/tests/controllers/docs/test_docs.py
```diff
@@ -225,3 +227,145 @@ def test_only_docs_browser_friendly(
...
+    # mismatch) rather than silently disappearing from the documentation
+    for method, path in PUBLIC_OPERATIONS:
+        assert schema["paths"][path][method].get("security") == []
+    for method, path in USER_ONLY_OVERRIDES:
+        assert schema["paths"][path][method].get("security") == SESSION_ONLY_SECURITY
+
+    # Operations in none of the groups must inherit the root-level
+    # requirements and must not opt out of them
+    assert "security" not in schema["paths"]["/v1/artefacts/{artefact_id}"]["get"]
+    assert "security" not in schema["paths"]["/v1/version"]["get"]
```
wctaylor: Are these just arbitrary choices?

wctaylor (commented):
One minor question, but otherwise I think this looks like a welcome change! Out of curiosity, is the PR description still accurate, or have some of the feedback rounds tweaked the output?

backend/tests/controllers/docs/test_docs.py
astrojuanlu: Yes. I can either leave them as-is, or replace them by a loop that checks all the schema.

astrojuanlu:
Updated the PR description! Will rebase on top of current `main`.

backend/tests/controllers/docs/test_docs.py
wctaylor: Might be overkill, but I lean towards the loop. Alternatively, just have the `/v1/version` check. The choice of `/v1/artefacts/{artefact_id}` without other, similar endpoints just makes it feel extra arbitrary, while the `/v1/version` one alone is clearly its own thing and which is likely to represent a case with no special handling

astrojuanlu: Done! Implemented a loop.

wctaylor (approved):
Thanks for the contribution!

